PR #14, opened 2020-07-02
Create Job JS

riya-n:
This includes the JS function that makes a POST request to create a job, given the user's inputs. The only validation check it includes is for max >= min (for job pay).


walk-in-interview/src/main/webapp/new-job.html
```diff
@@ -48,7 +48,7 @@
             <div class="job-field">
                 Job Duration (optional):
                 <select id="new-job-duration" name="job-duration">
-                    <option selected value="">Select</option>
+                    <option selected value="">Other</option>
```
SibingWu: What do u mean by "Other"?

riya-n: I thought this might be more appropriate than the user leaving the job duration option as "select". Using "other" indicates a non specified job duration, which I think was our intention when we made job duration optional.

walk-in-interview/src/main/webapp/new-job.html
```diff
@@ -4,22 +4,22 @@
...
     <meta charset="UTF-8">
     <link rel="stylesheet" href="style.css">
-    <script src="script.js"></script>
+    <script src="new-job.js"></script>
 </head>
 
 <body>
 
-    <form>
+    <form id="new-job-form" onsubmit="addJob()">
```
yohanlaunay: ah the return of submit. does that mean that your POST API will return html instead of json?

walk-in-interview/src/main/webapp/new-job.html
```diff
@@ -60,7 +60,7 @@
 
             <div class="job-field">
                 Post Expiry:
-                <input type="date" id="new-post-expiry" name="post-expiry" 
+                <input type="date" id="new-job-expiry" name="job-expiry" 
                     min="2020-07-01" max="2021-06-30" required>
                 <!-- TODO(issue/11) -->
```
yohanlaunay: nit, please indicate some details about the issue not just the issue Id. something like `TODO(issue/11) do this thing -->`

walk-in-interview/src/main/webapp/new-job.html
```diff
@@ -48,7 +48,7 @@
             <div class="job-field">
                 Job Duration (optional):
                 <select id="new-job-duration" name="job-duration">
```
yohanlaunay: i wonder why job duration is important maybe i'm misremembering from the doc.

walk-in-interview/src/main/webapp/new-job.js
```diff
@@ -0,0 +1,70 @@
...
+  });
+
+  const expiry = formElements[NEW_JOB_EXPIRY_ID].value;
+  const duration = formElements[NEW_JOB_DURATION_ID].value;
+
+  const jobDetails = {
+    jobName: name,
+    jobLocation: {
+      address: address,
+      lat: 1.3039, // TODO(issue/13)
```
yohanlaunay: nit: indicate more details about the issue

walk-in-interview/src/main/webapp/new-job.js
```diff
@@ -0,0 +1,70 @@
...
+    jobPay: {
+      frequency: payFrequency,
+      min: payMin,
+      max: payMax,
+    },
+    requirements: requirementsList,
+    postExpiry: expiry,
+  };
+
+  duration !== '' ? jobDetails.jobDuration = duration : '';
```
yohanlaunay: err i think you inverted your things here, the '' is going nowhere. You probably want either an if OR you want to clear the jobDetails.jobDuration.

yohanlaunay (commented):
don't checkin empty files (script.js).

yohanlaunay (approved):

walk-in-interview/src/main/webapp/new-job.html
```diff
@@ -4,30 +4,30 @@
 <head>
     <meta charset="UTF-8">
     <link rel="stylesheet" href="style.css">
-    <script src="script.js"></script>
+    <script src="new-job.js"></script>
```
mebjas: super nit, `js` files are usually added after the body part so that the rendering happens first and then dynamic js runs.

walk-in-interview/src/main/webapp/new-job.js
```diff
@@ -0,0 +1,74 @@
+
```
mebjas, line 6: nit, add a small comment which suggest what this file does.

SibingWu (approved):

walk-in-interview/src/main/webapp/new-job.js
```diff
@@ -0,0 +1,74 @@
...
+};
+const REQUIREMENTS_LIST_ID = 'requirements-list';
+const NEW_JOB_EXPIRY_ID = 'new-job-expiry';
+const NEW_JOB_DURATION_ID = 'new-job-duration';
+
+/**
+ * Adds a new job listing by making a POST request to the /jobs servlet.
+ * @return {boolean} If the request given the parameters was valid or not.
+ */
+function addJob() {
```
mebjas: Optional, you don't really need a form here you can attach an event listener to the submit button and then perform the same operations as in the end the request is made async using AJAX.

Is there a specific reason to use `<form>` here?

walk-in-interview/src/main/webapp/new-job.js
```diff
@@ -0,0 +1,74 @@
...
+  event.preventDefault();
```
mebjas: event is not defined in the arguments, `event` should be `undefined` here IIUC

this would work if your method signature has event argument
```
function addJon(event) {
    event.preventDefault();
}
```

And `Form#onSubmit` sends the event object.

walk-in-interview/src/main/webapp/new-job.js
```diff
@@ -0,0 +1,74 @@
...
+  if (duration !== '') {
+    jobDetails.jobDuration = duration;
+  }
+
+  fetch('/jobs', {
+    method: 'POST',
+    headers: {'Content-Type': 'application/json'},
+    body: JSON.stringify(jobDetails),
+  }).then((response) => response.text()).then((data) => {
+    return data.status === 200;
```
mebjas: no one is consuming the response of this promise, what's the intent here?

walk-in-interview/src/main/webapp/new-job.js
mebjas: If you use the event listener approach then your `addJob` can return a `Promise` as well and you can consume it like

```js
function addJob() {} // returns a Promise

const submitButton = document.getElementById('submit_button_id');
submitButton. addEventListener('click', _ => {
    addJob()
    .then(ignore -> { 
        // on success things
    })
    .catch(failure -> {
        // failure cases
    });
});
```

mebjas: One more suggestion you can define Job as a class and `ECMA5` is supported in all major browsers now so something like

```js
class Job {
    constructor(jobInput) {}

    // async functions are similar to Promise, alternatively use submit() which returns a Promise
    async submitAsync() {}
    
    // factory method which could abstract getting data from DOM and creating job object
    static createFromUserInputs() {
        // get everything from DOM
        return new Job(jobInput);
    } 
}
```

mebjas (changes requested):

mebjas (approved):

mebjas:
Marking my comments as optional, `Approval`

walk-in-interview/src/main/webapp/new-job.js
riya-n: I do believe that if this function returns false, then the form submission doesn't fully go through and so the user isn't redirected to the next page (after submitting the form). This could happen if the user has entered invalid fields or if the new job object could not be created/added to the database for some reason.

walk-in-interview/src/main/webapp/new-job.js
riya-n: It seems that the event object is defined even without passing it in the arguments. I tried passing it as an argument but then when I did event.preventDefault(), it didn't work. Do you know why this might be?

walk-in-interview/src/main/webapp/new-job.html
```diff
@@ -4,69 +4,70 @@
...
+    </div>
+
 </body>
 
-</html>
+</script>
+<script src="new-job.js"></script>
+</script>
+
+</html>
```
mebjas: nit, newline at end of each file.

walk-in-interview/src/main/webapp/new-job.js
```diff
@@ -0,0 +1,94 @@
...
+function addJob() {
+  const name = document.getElementById(NEW_JOB_TITLE_ID).value;
+  const address = document.getElementById(NEW_JOB_ADDRESS_ID).value;
+  const description = document.getElementById(NEW_JOB_DESCRIPTION_ID).value;
+  const payFrequency = document.getElementById(NEW_JOB_PAY.FREQUENCY_ID).value;
+
+  const payMin = document.getElementById(NEW_JOB_PAY.MIN_ID).valueAsNumber;
+  const payMax = document.getElementById(NEW_JOB_PAY.MAX_ID).valueAsNumber;
+  if (payMin > payMax) {
+    return Promise.reject(new Error('min greater than max'));
```
mebjas: `Promise` is used for async stuff like a network call example `fetch` method below. I'd recommend separating non async tasks as the two methods with different responsibilities

```js
/** Gets job detail from user input. */
function getJobDetailsFromUserInput() {}

/** Function validate input */
function validateUserInput() {}
```

Now consume it like
```js
const submitButton = document.getElementById('new-job-submit');
submitButton. addEventListener('click', (_) => {
  if (!validateUserInput()) {
    // show some error UI
    return;
  }
 
  const jobDetails = getJobDetailsFromUserInput();
  fetch('/jobs', {
    method: 'POST',
    headers: {'Content-Type': 'application/json'},
    body: JSON.stringify(jobDetails),
   })
   .then((repsonse) => repsonse.text())
   .then((data) => {
      console.log('data', data);
      window.location.href= HOMEPAGE_PATH;
    })
    .catch((error) => {
      // some error ui
      console.log('error', error);
    });
});


```

mebjas (approved):

walk-in-interview/src/main/webapp/new-job.js
```diff
@@ -0,0 +1,124 @@
...
+      .then((repsonse) => repsonse.text())
+      .then((data) => {
+        console.log('data', data);
+        window.location.href= HOMEPAGE_PATH;
+      })
+      .catch((error) => {
+        document.getElementById(NEW_JOB_ERROR_MESSAGE.ID).innerText =
+          NEW_JOB_ERROR_MESSAGE.ERROR_RESPONSE;
+        console.log('error', error);
+      });
```
mebjas, line 123: nit, this indentation looks off, shouldn't it be one tab backwards?

mebjas (approved):

walk-in-interview/src/main/webapp/new-job.js
riya-n: eslint is complaining for any other indentation :/

PR #20, opened 2020-07-03
Add create job page elements dynamically

riya-n:
This is to add the options for the select elements dynamically rather than having them hardcoded into the html.


walk-in-interview/src/main/webapp/new-job.js
```diff
@@ -7,21 +7,113 @@
...
+const NEW_JOB_DURATION_ID = 'new-job-duration';
+const REQUIREMENTS_LIST_NAME = 'requirements-list';
+const NEW_JOB_SUBMIT_ID = 'new-job-submit';
+const NEW_JOB_CANCEL_ID = 'new-job-cancel';
+
+const HOMEPAGE_PATH = '/walk-in-interview/src/main/webapp/index.html';
+
+/**
+ * Fields related to job pay.
+ * @enum {stirng}
```
htthach: Nit: typo "string"? Btw, is this enum?

walk-in-interview/src/main/webapp/new-job.js
```diff
@@ -7,21 +7,113 @@
...
   MIN_ID: 'new-job-pay-min',
   MAX_ID: 'new-job-pay-max',
 };
-const REQUIREMENTS_LIST_NAME = 'requirements-list';
-const NEW_JOB_EXPIRY_ID = 'new-job-expiry';
-const NEW_JOB_DURATION_ID = 'new-job-duration';
+
+/**
+ * Options for job duration.
+ * @enum {stirng}
```
htthach: Nit: typo? Same for below

walk-in-interview/src/main/webapp/new-job.js
```diff
@@ -7,21 +7,113 @@
...
+function addRequirementsList() {
+  // TODO(issue/17): get from the database and render dynamically
+}
+
+/** Dynamically add the options for job pay frequency. */
+function addJobPayFrequencyOptions() {
+  const jobPaySelect = document.getElementById(NEW_JOB_PAY.FREQUENCY_ID);
+  jobPaySelect.options.length = 0;
+
+  (jobPaySelect.options[0] = new Option('Select', ''))
```
htthach: Should this be two statements?

walk-in-interview/src/main/webapp/new-job.js
```diff
@@ -114,6 +206,7 @@ submitButton. addEventListener('click', (_) => {
       .then((repsonse) => repsonse.text())
       .then((data) => {
         console.log('data', data);
+        document.getElementById(NEW_JOB_ERROR_MESSAGE.ID).innerText = '';
```
htthach: when there is error below, should we also set this element text?

walk-in-interview/src/main/webapp/script.js
```diff
@@ -0,0 +1 @@
+
```
htthach: Accidental newline?

walk-in-interview/src/main/webapp/new-job.js
```diff
@@ -7,21 +7,113 @@
...
+const JOB_PAY_FREQUENCY = {
+  HOURLY: 'Hourly',
+  WEEKLY: 'Weekly',
+  MONTHLY: 'Monthly',
+  YEARLY: 'Yearly',
+};
+
+/**
+ * Error messages.
+ * @enum {stirng}
```
htthach: this is not really enum right?

walk-in-interview/src/main/webapp/new-job.js
riya-n: I thought I already do this?

walk-in-interview/src/main/webapp/new-job.js
```diff
@@ -122,3 +211,8 @@ submitButton. addEventListener('click', (_) => {
         console.log('error', error);
       });
 });
+
+const cancelButton = document.getElementById(NEW_JOB_CANCEL_ID);
+cancelButton. addEventListener('click', (_) => {
```
SibingWu: here seems a redundant space after the dot

SibingWu (commented):
I am not sure if we need to extract all text in the html out and put into a map, to make the string with different language version, although so far we maybe only have English version

walk-in-interview/src/main/webapp/new-job.html
```diff
@@ -9,66 +9,49 @@
...
+                <label for="o-level">O Levels</label>
+                <input type="checkbox" id="drivers-license" name="requirements-list" value="drivers-license">
+                <label for="drivers-license">Driver's License</label>
+                <!-- TODO(issue/17): render based on list in database -->
+            </div>
+
+            <div class="job-field">
+                <div id="new-job-pay-title"></div>
+                <select id="new-job-pay-frequency"></select>
+                <input id="new-job-pay-min">
```
yohanlaunay: you probably want to wrap your input for min and max into a div so you can more easily control the alignment compared to the select and the title

walk-in-interview/src/main/webapp/new-job.html
```diff
@@ -9,66 +9,49 @@
...
+            <textarea class="job-field" id="new-job-description"></textarea>
+            <input class="job-field" id="new-job-address" />
+
+            <div class="job-field">
+                <div id="new-job-requirements-title"></div>
+                <input type="checkbox" id="o-level" name="requirements-list" value="o-level">
+                <label for="o-level">O Levels</label>
+                <input type="checkbox" id="drivers-license" name="requirements-list" value="drivers-license">
+                <label for="drivers-license">Driver's License</label>
+                <!-- TODO(issue/17): render based on list in database -->
```
yohanlaunay: i actually recommend you render list based on javascript first. then your javascript fetches the list from database.

walk-in-interview/src/main/webapp/new-job.js
```diff
@@ -4,24 +4,214 @@
...
+const NEW_JOB_SUBMIT_ID = 'new-job-submit';
+const NEW_JOB_CANCEL_ID = 'new-job-cancel';
+
+const NEW_JOB_PAGE_TITLE_ID = 'new-job-page-title';
+const NEW_JOB_REQUIREMENTS_TITLE_ID = 'new-job-requirements-title';
+const NEW_JOB_PAY_TITLE_ID = 'new-job-pay-title';
+const NEW_JOB_DURATION_TITLE_ID = 'new-job-duration-title';
+const NEW_JOB_EXPIRY_TITLE_ID = 'new-job-expiry-title';
+
+const HOMEPAGE_PATH = '/walk-in-interview/src/main/webapp/index.html';
```
yohanlaunay: are you sure the path is correct? i thought js path should be starting from the index not the src root

walk-in-interview/src/main/webapp/new-job.js
```diff
@@ -4,24 +4,214 @@
  * is submitted.
  */
 
+/**
+ * Titles for all the sections/fields in different languages.
+ */
+const NEW_JOB_PAGE_TITLE = {
```
yohanlaunay: i think you can move all these localized strings into a strings.js if you want to do it per page, you can do `new-job/strings.en.js`

And have something like:
```
const AppStrings = AppStrings || {}
AppStrings['new-job'] = {
   'page-title': 'New Job Post',
    [...]
}
```

walk-in-interview/src/main/webapp/new-job.js
```diff
@@ -4,24 +4,214 @@
...
   MAX_ID: 'new-job-pay-max',
 };
-const REQUIREMENTS_LIST_NAME = 'requirements-list';
-const NEW_JOB_EXPIRY_ID = 'new-job-expiry';
-const NEW_JOB_DURATION_ID = 'new-job-duration';
+
+/**
+ * Options for job duration.
+ */
+const JOB_DURATION = {
```
yohanlaunay: This should be strings (localized)

walk-in-interview/src/main/webapp/new-job.js
```diff
@@ -4,24 +4,214 @@
...
+  TWO_WEEKS: '2 Weeks',
+  ONE_MONTH: '1 Month',
+  SIX_MONTHS: '6 Months',
+  ONE_YEAR: '1 Year',
+};
+
+/**
+ * Options for frequency of job pay.
+ */
+const JOB_PAY_FREQUENCY = {
```
yohanlaunay: This should be localized strings

walk-in-interview/src/main/webapp/new-job.js
```diff
@@ -4,24 +4,214 @@
...
+  HOURLY: 'Hourly',
+  WEEKLY: 'Weekly',
+  MONTHLY: 'Monthly',
+  YEARLY: 'Yearly',
+};
+
+/**
+ * Error messages.
+ */
 const NEW_JOB_ERROR_MESSAGE = {
```
yohanlaunay: i recommend you don't mix strings and non-strings.

walk-in-interview/src/main/webapp/new-job/new-job.html
```diff
@@ -0,0 +1,55 @@
+<!DOCTYPE html>
```
yohanlaunay: should the file name be index.html?

walk-in-interview/src/main/webapp/new-job/new-job.html
```diff
@@ -0,0 +1,55 @@
...
+            </div>
+        </div>
+
+        <div class="error-message" id="new-job-error-message"></div>
+
+</body>
+
+<script type="module" src="./new-job.js"></script>
+
+</html>
```
yohanlaunay: nit: newline at end of file

walk-in-interview/src/main/webapp/new-job/new-job.js
```diff
@@ -0,0 +1,293 @@
...
+const NEW_JOB_STRINGS = AppStrings['new-job'];
+const HOMEPAGE_PATH = '../index.html';
+
+window.onload = () => {
+  addJobPageElements();
+};
+
+/** Adds all the titles to the fields on this page. */
+function addJobPageElements() {
+  const cancelButton = document.getElementById(NEW_JOB_STRINGS['cancel'].id);
```
yohanlaunay: as mentioned in previous comments, the strings should not have ids like that, just strings. your ID can be somewhere else and reference the string id.

For example something like that:
```
const CANCEL_BUTTTON = {
    id: 'btn-cancel',
    label: STRINGS['btn_cancel'],
};
```

But i think even that is overkill. You can put the button id directly in your getElementByid:

```
const cancelButton = document.getElementById('btn-cancel');
cancelButton.setAttribute('value', STRINGS['btn-cancel']);
cancelButton.setAttribute('type', 'reset');
```

Defining constants for all the HTML elements will become really tedious specially when they are only used twice: once in the html and once in the javascript.

OR you use the same ID for the button everywhere: in your strings, in your html and in your javascript so you only have a single string constant:
```
const CANCEL_BUTTON_ID = 'btn-cancel';
[...]
const cancelButton = document.getElementById(CANCEL_BUTTON_ID);
cancelButton.setAttribute('value', STRINGS[CANCEL_BUTTON_ID]);
cancelButton.setAttribute('type', 'reset');
```

walk-in-interview/src/main/webapp/new-job/new-job.js
```diff
@@ -0,0 +1,293 @@
...
+
+/** Add the list of requirements that are stored in the database. */
+function addRequirementsList() {
+  const requirementsList = getRequirementsList();
+  const requirementsListElement =
+    document.getElementById(NEW_JOB_STRINGS['job-requirements-list'].id);
+
+  requirementsListElement.innerHTML = '';
+  for (const key in requirementsList) {
+    if ({}.hasOwnProperty.call(requirementsList, key)) {
```
yohanlaunay: i'm surprised you cannot do
```
if( requirementsList.hasOwnProperty(key) )
```

Also can you use a guard clause instead to reduce indentation:
```
if( ! requirementsList.hasOwnProperty(key) ){
    continue;
}
```

walk-in-interview/src/main/webapp/new-job/new-job.js
```diff
@@ -0,0 +1,293 @@
...
+      const requirementElement = document.createElement('li');
```
yohanlaunay: FYI, one thing i used to like doing instead of this was to have a bunch of templates in the HTML directly:
basically you have a hidden div called "templates" at the bottom of your page that contains all the dynamically created items. then in javascript you load the template you want (by id), clone it, remove the id, and fill it up.

walk-in-interview/src/main/webapp/new-job/new-job.js
```diff
@@ -0,0 +1,293 @@
+/**
+ * This file is specific to new-job.html. It renders the fields on the
+ * page dynamically, and it also makes the POST request when the form
+ * is submitted.
+ */
+
+import {AppStrings} from './strings.en.js';
```
yohanlaunay: super nit:
```
const CurrentLocale = 'en'; <-- then derive from there. Ideally should come from the browser information or cookies/local information across your app.

import {AppStrings} from "./strings.{CurrentLocale}.js"; <-- something like that (i don't remember the syntax).

```

walk-in-interview/src/main/webapp/new-job/new-job.js
```diff
@@ -0,0 +1,293 @@
...
+      frequency: payFrequency,
+      min: payMin,
+      max: payMax,
+    },
+    requirements: requirementsList,
+    postExpiry: expiry,
+  };
+
+  if (duration !== '') {
+    jobDetails.jobDuration = duration;
```
yohanlaunay: so that means that jobduration will not be present in your javascript if it's empty. I suppose the field is optional and that's intended.

walk-in-interview/src/main/webapp/new-job/new-job.js
```diff
@@ -0,0 +1,293 @@
+/**
+ * This file is specific to new-job.html. It renders the fields on the
+ * page dynamically, and it also makes the POST request when the form
+ * is submitted.
+ */
+
+import {AppStrings} from './strings.en.js';
+
+const NEW_JOB_STRINGS = AppStrings['new-job'];
```
yohanlaunay: FYI you only need NEW_JOB_STRINGS there, so maybe just call it STRINGS? and if you need non 'new-job' strings (ex: you have common ones like product name and such) then you can have a COMMON_STRINGS ?

walk-in-interview/src/main/webapp/new-job/new-job.js
riya-n: yep

riya-n: Although it seems Sibing is having some trouble using Optionals with Firestore, so this might be changed later.

yohanlaunay (approved):

walk-in-interview/src/main/webapp/new-job/script.js
```diff
@@ -0,0 +1,296 @@
...
+  }
+
+  return false;
+}
+
+const submitButton = document.getElementById('new-job-submit');
+submitButton.addEventListener('click', (_) => {
+  if (!validateRequiredUserInput()) {
+    document.getElementById('new-job-error-message').innerText =
+      STRINGS['new-job-error-message'].message;
```
htthach: Maybe validateRequiredUserInput some how returns an error message indicating which field is missing/ wrongly entered? And we show it here? If no error message, the input is correct? 
Usually it's not trivial and friendly when a form just says "there's an error" without pointing out where or why.

htthach (approved):

walk-in-interview/src/main/webapp/new-job/index.html
```diff
@@ -0,0 +1,62 @@
...
+    <meta charset="UTF-8">
+    <link rel="stylesheet" href="../style.css">
+</head>
+
+<body>
+
+    <div class="header-container">
+        <input id="new-job-cancel">
+        <div id="new-job-page-title"></div>
+        <input id="new-job-submit">
```
SibingWu, line 17: if the type not specified, seems that the default is `text`?

walk-in-interview/src/main/webapp/new-job/index.html
```diff
@@ -0,0 +1,62 @@
...
+
+    <div class="content-container">
+        <div class="content-container">
+            <input class="job-field" id="new-job-title" />
+            <textarea class="job-field" id="new-job-description"></textarea>
+            <input class="job-field" id="new-job-address" />
+
+            <div class="job-field">
+                <div id="new-job-requirements-title"></div>
+                <ul id="new-job-requirements-list"></ul>
```
SibingWu, line 28: maybe add a TODO here to fill in the requirement list in the future

walk-in-interview/src/main/webapp/new-job/script.js
```diff
@@ -0,0 +1,296 @@
+/**
+ * This file is specific to new-job.html. It renders the fields on the
```
SibingWu, line 2: seems that "new-job.html" is deleted?

walk-in-interview/src/main/webapp/new-job/index.html
SibingWu, line 17: maybe u can add a comment to say that the content will be dynamically rendered in other js file

walk-in-interview/src/main/webapp/new-job/script.js
SibingWu: Agree

walk-in-interview/src/main/webapp/new-job/index.html
riya-n: I specify the type in JS and dynamically add it to this element

walk-in-interview/src/main/webapp/new-job/index.html
riya-n: This is currently being rendered from the JS side so I've added the TODO there since there is no more work to be done in the HTML file.

walk-in-interview/src/main/webapp/new-job/script.js
riya-n: I called it index.html instead, and put it in a new-job folder.

Issue #30, opened 2020-07-09
Add feature to delete a job post


SibingWu:
As per title
Change the job post status to DELETED instead of truly deleting in the database
So far backend section finishes. Front end section has not started yet.

PR #50, opened 2020-07-15
Add the feature to delete a job post - backend

SibingWu:
Related issue: #30 (not fixed yet, it is only the backend part)

* Add the servlet to change the status of a job post, which includes the feature to delete a job post
* Write test for `markJobPostAsDeleted`
* Only cover the backend part


walk-in-interview/src/main/java/com/google/job/servlets/ChangeJobStatusServlet.java
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    /** Handles the PATCH request from client. */
+    public void doPatch(HttpServletRequest request, HttpServletResponse response) {
+        try {
+            // Gets the target job post id
+            String jobId = ServletUtils.getStringParameter(request, JOB_ID_FIELD, /* defaultValue= */ "");
+
+            // Gets the target status
+            JobStatus jobStatus = getTargetStatus(request);
```
htthach: Maybe we should not let frontend control this status by directly modifying the field? Maybe frontend indicate the desire "action" instead? 
E.g Instead of frontend setting the new status as deleted, we allow frontend to send a DELETE http request, on server, we handle in doDelete(). Inside doDelete() we mark status as DELETED or whatever status we desire as it's internal logic of the backend? Maybe in the future, we can even change it to "INACTIVE" and it won't affect frontend?

walk-in-interview/src/main/java/com/google/job/servlets/ChangeJobStatusServlet.java
```diff
@@ -0,0 +1,101 @@
...
+    public void init() {
+        this.jobsDatabase = new JobsDatabase();
+    }
+
+    @Override
+    public void service(HttpServletRequest request, HttpServletResponse response) throws ServletException, IOException {
+        // Explicitly routes PATCH requests to a doPatch method since by default HttpServlet doesn't do it for us.
+        if (request.getMethod().equalsIgnoreCase(PATCH_METHOD_TYPE)){
+            doPatch(request, response);
+        } else {
```
htthach: Maybe we also support "Delete" method type?

walk-in-interview/src/main/java/com/google/job/servlets/ChangeJobStatusServlet.java
```diff
@@ -0,0 +1,101 @@
...
+
+            // Verifies if the current user can update the job post with this job id.
+            // TODO(issue/25): incorporate the account stuff into job post.
+            verifyUserCanUpdateJob(jobId);
+
+            // Change the status
+            switch (jobStatus) {
+                case ACTIVE:
+                case EXPIRED:
+                    throw new UnsupportedOperationException("Not implemented");
```
riya-n: nit: add the TODO with the issue here too

walk-in-interview/src/main/java/com/google/utils/ServletUtils.java
```diff
@@ -102,6 +103,22 @@ public static Date getDateParameter(HttpServletRequest request,
         }
     }
 
+    /**
+     * @return the value of parameter with the {@code name} in the {@code request}
```
riya-n: is there no need to write @param in the docs? same comment on docs for other functions

SibingWu: this format is also allowed

riya-n (approved):

walk-in-interview/src/main/java/com/google/job/servlets/ChangeJobStatusServlet.java
SibingWu: As discussed offline, change to a new end point "/jobs/delete"

htthach (approved):

walk-in-interview/src/main/java/com/google/job/data/Job.java
```diff
@@ -273,4 +273,12 @@ public int hashCode() {
 
         return hashCode;
     }
+
+    @Override
+    public String toString() {
+        return String.format("Job{jobId=%s, jobStatus=%s, jobTitle=%s, jobLocation=%s, "
```
yohanlaunay, line 279: i'm surprised the code doesn't mind you using %s even for integers/longs

walk-in-interview/src/main/java/com/google/job/data/JobsDatabase.java
```diff
@@ -59,6 +59,17 @@
...
 
+    /**
+     * Marks a job post as DELETED.
+     *
+     * @param jobId Cloud Firestore Id of the job post.
+     * @return A future of the detailed information of the update.
+     */
+    public Future<WriteResult> markJobPostAsDeleted(String jobId) {
+        DocumentReference documentReference = FireStoreUtils.getFireStore().collection(JOB_COLLECTION).document(jobId);
+        return documentReference.update(JOB_STATUS_FIELD, JobStatus.DELETED);
```
yohanlaunay: I actually think this should be done in a transaction. right now your database class is delegating the existence check for jobId to the caller. Typically you would do a transaction on document(jobId), if the document doesn't exist you should fail the transaction or something, if it exists you can update the status field.
https://firebase.google.com/docs/firestore/manage-data/transactions

This same comment is valid for anything mutating a job by Id. You start a transaction, verify that the job exists, then do the mutation (or fail).

walk-in-interview/src/main/java/com/google/job/servlets/MarkJobDeleteServlet.java
```diff
@@ -0,0 +1,84 @@
...
+
+    /** Handles the PATCH request from client. */
+    public void doPatch(HttpServletRequest request, HttpServletResponse response) {
+        try {
+            // Gets the target job post id
+            String jobId = ServletUtils.getStringParameter(request, JOB_ID_FIELD, /* defaultValue= */ "");
+
+            // Verifies if the current user can update the job post with this job id.
+            // TODO(issue/25): incorporate the account stuff into job post.
+            verifyUserCanUpdateJob(jobId);
```
yohanlaunay: this should not be needed. this should be done by jobsDb

walk-in-interview/src/main/java/com/google/job/servlets/MarkJobDeleteServlet.java
```diff
@@ -0,0 +1,84 @@
...
+
+            // Changes the status to DELETED
+            this.jobsDatabase.markJobPostAsDeleted(jobId);
+
+            // TODO(issue/51): after "interest list" is involved,
+            //  remove the DELETED job post from applicant's interest list
```
yohanlaunay: FYI you don't need to remove the job from the interest list, just filter it out on the server side when you retrieve the list of favorited jobs

walk-in-interview/src/main/java/com/google/job/data/JobsDatabase.java
```diff
@@ -44,19 +49,64 @@
...
-                .collection(JOB_COLLECTION).document(jobId)
-                .set(job);
+        // Runs an asynchronous transaction
+        ApiFuture<DocumentReference> futureTransaction = FireStoreUtils.getFireStore().runTransaction(transaction -> {
+            final DocumentReference documentReference = FireStoreUtils.getFireStore()
+                    .collection(JOB_COLLECTION).document(jobId);
+
+            // Verifies if the current user can update the job post with this job id
+            // TODO(issue/25): incorporate the account stuff into job post.
+            verifyUserCanUpdateJob(jobId);
```
yohanlaunay: this doesn't really work. You should access your db within the transaction at this time, see the link i gave you:
https://firebase.google.com/docs/firestore/manage-data/transactions
```
 DocumentSnapshot snapshot = transaction.get(docReference);
 if( snapshot == null ){ [...] // job doesn't exist }
 [...]
```
Same for the other one.

yohanlaunay (approved):

PR #54, opened 2020-07-16
Extract common js method

SibingWu:
As per title, 
* extracts out some commonly used job-related js function from new-job/script.js so that it can be imported and reused
* merges into only one strings.en.js


walk-in-interview/src/main/webapp/common-function.js
```diff
@@ -0,0 +1,140 @@
...
+      'LANGUAGE_ENGLISH': 'English',
+      'DRIVING_LICENSE_C': 'Category C Driving License',
+    };
+}
+
+/**
+ * Gets job detail from user input.
+ * @return {Object} containing the user inputs.
+ */
+function getJobDetailsFromUserInput() {
```
riya-n: do we need to pass `document` in as a param?

walk-in-interview/src/main/webapp/common-functions.js
```diff
@@ -0,0 +1,140 @@
...
+      'LANGUAGE_ENGLISH': 'English',
+      'DRIVING_LICENSE_C': 'Category C Driving License',
+    };
+}
+
+/**
+ * Gets job detail from user input.
+ * @return {Object} containing the user inputs.
+ */
+function getJobDetailsFromUserInput() {
```
riya-n: do we need to pass `document` in as a param?

walk-in-interview/src/main/webapp/new-job/script.js
```diff
@@ -13,6 +13,8 @@ const CurrentLocale = 'en';
  */
 import {AppStrings} from './strings.en.js';
 
+import {getRequirementsList, getJobDetailsFromUserInput, validateRequiredUserInput, setErrorMessage} from "../common-function.js";
```
riya-n: from "../common-functions.js"

riya-n (approved):

riya-n (commented):
Maybe the `strings.en.js` file can also be moved outside the new-job folder since many of the strings would be similar.

walk-in-interview/src/main/webapp/common-functions.js
```diff
@@ -0,0 +1,140 @@
...
+    const address = document.getElementById('address');
+    const postalCode = document.getElementById('postal-code');
+    const payFrequency = document.getElementById('pay-frequency').value;
+    const payMin = document.getElementById('pay-min').valueAsNumber;
+    const payMax = document.getElementById('pay-max').valueAsNumber;
+    const duration = document.getElementById('duration').value;
+    const expiry = document.getElementById('expiry').valueAsNumber;
+
+    if (name.value === '') {
+        setErrorMessage(/* msg */ name.placeholder,
```
riya-n: you changed the setErrorMessage params. There should be a third param here for the `defaultMessage`. I would suggest that you leave the setErrorMessage function as it was before, and import `string.en.js` into this file.
```
function setErrorMessage(msg, includesDefault) {
  document.getElementById('error-message').innerText =
    (includesDefault ? STRINGS['error-message'] + msg : msg);
}
```

walk-in-interview/src/main/webapp/common-functions.js
SibingWu: according to experiment, no need

walk-in-interview/src/main/webapp/common-functions.js
riya-n: or you could simply have a constant for `STRINGS['error-message']` instead of importing it. i.e. `const DEFAULT_ERROR_MSG = 'xxx';`

walk-in-interview/src/main/webapp/strings.en.js
```diff
@@ -9,14 +9,20 @@
  */
 export const AppStrings = {};
 
+/**
+ * Common strings used in all pages.
+ */
+AppStrings['constant'] = {
```
riya-n: maybe don't call this `constant` and instead maybe `common` is better?

walk-in-interview/src/main/webapp/common-functions.js
```diff
@@ -0,0 +1,41 @@
...
+    };
+}
+
+/**
+ * Sets the error message according to the param.
+ * @param {String} msg the message that the error div should display.
+ * @param {boolean} includesDefault whether the default.
+ * message should be included.
+ */
+function setErrorMessage(msg, includesDefault) {
```
yohanlaunay: can you pass the errorMessageElementId to the method instead of using the hardcoded one?

yohanlaunay: FYI `includesDefault=true` should work afaik.

walk-in-interview/src/main/webapp/new-job/script.js
```diff
@@ -253,46 +229,46 @@ function validateRequiredUserInput() {
...
+          .textContent, /** includes default msg */ true);
+      return false;
   }
 
   if (Number.isNaN(expiry)) {
-    setErrorMessage(/* msg */ document.getElementById('expiry-title')
-        .textContent, /** includes default msg */ true);
-    return false;
+      setErrorMessage(/* msg */ document.getElementById('expiry-title')
+          .textContent, /** includes default msg */ true);
```
yohanlaunay: FYI annotations look like this: `/* paramName= */true`. You need the `=` and you don't want the double ** which is used for javadoc :) But as i mentioned in the common file you might be able to have default value.

walk-in-interview/src/main/webapp/strings.en.js
```diff
@@ -37,3 +43,37 @@ AppStrings['homepage'] = {
   'filter-by-submit': 'Apply Filters',
   'job-listings-title': 'Job Listings:',
 };
+
+/**
+ * All the strings to be used in the new-job folder.
+ */
+AppStrings['new-job'] = {
```
yohanlaunay, line 50: any reason why you want all the strings to be in one place? shouldn't new-job strings be inside the new-job package?

SibingWu: later it in update-job feature, many strings are the same, so simply merge into one

yohanlaunay (approved):

PR #61, opened 2020-07-20
Add GET jobs/listings servlet

riya-n:
This makes a GET request to get all job listings based on the filters on the homepage: sort by (salary ascending/descending), show by region (Singapore regions), filter by (min/max limits). For reference, this is what the UI currently looks like:
image
Note that this PR does not include pagination for the job listings on the homepage, and also does not tests for the database function that retrieves the job listings.


walk-in-interview/src/main/java/com/google/job/data/JobPayment.java
```diff
@@ -5,20 +5,22 @@
...
     private final int max;
     private final PaymentFrequency paymentFrequency;
+    private final long annualMax;
 
     private volatile int hashCode;
 
     // For serialization
     public JobPayment() {
-        this(/* min= */0, /* max= */0, PaymentFrequency.HOURLY);
+        this(/* min= */0, /* max= */0, PaymentFrequency.HOURLY, /* annualMax */ 0);
```
yohanlaunay: super nit: `/* annualMax= */0`

walk-in-interview/src/main/java/com/google/job/data/JobPage.java
```diff
@@ -0,0 +1,106 @@
...
+
+    @Override
+    public String toString() {
+        return String.format("JobPage{jobList=%s, totalCount=%d, range=%s}",
+                jobList, totalCount, range);
+    }
+
+    private static void validateParameters(List<Job> jobList, long totalCount, Range<Integer> range)
+            throws IllegalArgumentException {
+        if (jobList == null || jobList.size() < 0) {
```
yohanlaunay: jobList and range are not nullable

walk-in-interview/src/main/java/com/google/job/data/JobPage.java
```diff
@@ -0,0 +1,106 @@
...
+public final class JobPage {
+    private final List<Job> jobList;
+    private final long totalCount;
+    private final Range<Integer> range;
+
+    private volatile int hashCode;
+
+    // For serialization
+    public JobPage() {
+        this(/* jobList */ new LinkedList(), /* totalCount */0,
```
yohanlaunay: nit: `ImmutableList.of()`

walk-in-interview/src/main/java/com/google/job/data/Filter.java
```diff
@@ -0,0 +1,36 @@
...
+    private final String filterId;
+
+    Filter(String filterId) {
+        this.filterId = filterId;
+    }
+
+    /**
+     * Returns the stable id representing the filter. Can be stored in database.
+     *
+     * @return the stable id.
```
yohanlaunay: super nit: `@return` is redundant here.

walk-in-interview/src/main/java/com/google/job/data/JobsDatabase.java
```diff
@@ -138,4 +146,59 @@
...
+     * @param order The ordering of the sorting.
+     * @param pageSize The number of job listings to be returned.
+     * @param pageIndex The page which we are on (pagination).
+     * @return Future of the JobPage object.
+     */
+    public static Future<JobPage> fetchJobPage(int minLimit, int maxLimit, SingaporeRegion region, Filter sortBy, Order order,
+        int pageSize, int pageIndex) throws IOException {
+        CollectionReference jobsCollection = FireStoreUtils.getFireStore().collection(JOB_COLLECTION);
+
+        if (!sortBy.equals(Filter.SALARY)) {
```
yohanlaunay: nit: TODO(issue/) support other filters.

walk-in-interview/src/main/java/com/google/job/data/JobsDatabase.java
```diff
@@ -138,4 +146,59 @@
...
+
+        // TODO(issue/34): add to the query to include pagination
+
+        ApiFuture<QuerySnapshot> snapshotFuture = query.get();
+
+        ApiFunction<QuerySnapshot, JobPage> jobFunction = new ApiFunction<QuerySnapshot, JobPage>() {
+            @NullableDecl
+            public JobPage apply(@NullableDecl QuerySnapshot snapshotFuture) {
+                List<QueryDocumentSnapshot> documents = snapshotFuture.getDocuments();
+                List<Job> jobList = new LinkedList<>();
```
yohanlaunay: nit: ImmutableList.Builder

walk-in-interview/src/main/java/com/google/job/data/JobsDatabase.java
```diff
@@ -138,4 +146,59 @@
...
+            .whereLessThanOrEqualTo(SALARY_FIELD, maxLimit)
+            .orderBy(SALARY_FIELD, Order.getQueryDirection(order));
+
+        if (!region.equals(SingaporeRegion.ENTIRE)) {
+            query = query.whereEqualTo(REGION_FIELD, region.name());
+        }
+
+        // TODO(issue/34): add to the query to include pagination
+
+        ApiFuture<QuerySnapshot> snapshotFuture = query.get();
```
yohanlaunay: nit: can you move `query.get()` and the jobFunction to the transform directly? I found it a bit difficult to figure out what you were doing.

```
return ApiFuture.transform(
   query.get(),
    (snapshotFuture) -> {
            [...]
      return new JobPage(...);
    }
);

```

walk-in-interview/src/main/java/com/google/job/data/Location.java
```diff
@@ -5,18 +5,20 @@
...
+    private final SingaporeRegion region;
     private final double latitude;
     private final double longitude;
 
     private volatile int hashCode;
 
     // For serialization
     public Location() {
-        this(/* address= */"", /* postalCode= */"dummy",
+        this(/* address= */"", /* postalCode= */"01dummy", /* region */ SingaporeRegion.CENTRAL,
```
yohanlaunay: nit: `/* region= */`

yohanlaunay: as discussed offline, you should revert your "01dummy" change back to "dummy" it should not matter what the value is as long as it's not empty. if this value matters, that means the gson deserialization is broken somehow

walk-in-interview/src/main/java/com/google/job/servlets/JobsListingsServlet.java
```diff
@@ -0,0 +1,221 @@
...
+     * @param sortBy The sorting of the list of jobs.
+     * @param order The ordering of the sorting.
+     * @param pageSize The number of job listings to be returned.
+     * @param pageIndex The page which we are on (pagination).
+     * @return JobPage object with all the details for the GET response.
+     */
+    private JobPage fetchJobPageDetails(int minSalary, int maxSalary, SingaporeRegion region,
+        Filter sortBy, Order order, int pageSize, int pageIndex) throws ServletException, ExecutionException, TimeoutException {
+        try {
+            return this.jobsDatabase.fetchJobPage(minSalary, maxSalary, region, sortBy, order, pageSize, pageIndex)
```
yohanlaunay: now that you have 7 arguments in there, you might want to consider passing a QueryBuilder object with default sensible values for order, pageSize, pageIndex.

walk-in-interview/src/main/java/com/google/job/servlets/JobsListingsServlet.java
```diff
@@ -0,0 +1,221 @@
...
+     * @throws IllegalArgumentException if the id is invalid.
+     */
+    private SingaporeRegion parseRegion(HttpServletRequest request) throws IllegalArgumentException {
+        String region = (String) request.getParameter(REGION_PARAM);
+
+        if (region == null || region.isEmpty()) {
+            throw new IllegalArgumentException("region param should not be null or empty");
+        }
+
+        return SingaporeRegion.getFromId(region); // IAE may be thrown
```
yohanlaunay: nit: 'IAE' <- IllegalArgumentException (same below)

walk-in-interview/src/main/webapp/common-functions.js
```diff
@@ -19,24 +19,26 @@ const STRINGS = AppStrings['common'];
...
 /**
  * Sets the error message according to the param.
- * @param {String} errorMessageElementId the element id for html error message div.
+ * @param {String} errorMessageElementId the element id
+ * for html error message div.
  * @param {String} msg the message that the error div should display.
- * @param {boolean} includesDefault whether the includes default message. Default to be true.
- * message should be included.
+ * @param {boolean} includesDefault whether the includes default message.
+ * Default to be true.
```
yohanlaunay: super nit: i think in terms of formatting this should be attached to the `@param`. So if you format automatically it should go to the new line with some space.

walk-in-interview/src/main/webapp/new-job/script.js
```diff
@@ -13,12 +13,24 @@ const CurrentLocale = 'en';
...
+const WEEKS_PER_YEAR = 52;
+const MONTHS_PER_YEAR = 12;
+
+/**
+ * Note that this is needed because in JS we can hold bigger Integer
+ * values than in Java.
+ */
+const JAVA_INTEGER_MAX_VALUE = Math.pow(2, 31) - 1;
+
+let currErrorField;
```
yohanlaunay: this is basically a global variable. Is it possible to add a class or a tag to the form field so that we don't need that variable?

walk-in-interview/src/main/webapp/new-job/script.js
```diff
@@ -179,43 +209,109 @@ function getJobDetailsFromUserInput() {
...
+  } else if ((digits >= 46 && digits <= 52) ||
+      digits === 81) {
+    return 'EAST';
+  } else if ((digits >= 72 && digits <= 73) ||
+      (digits >= 75 && digits <= 80)) {
+    return 'NORTH';
+  } else if ((digits >= 53 && digits <= 57) ||
+      digits === 82) {
+    return 'NORTH_EAST';
+  }
```
yohanlaunay, line 258: you are missing cases. for instance 74. I think you should throw instead of returning undef.

walk-in-interview/src/main/webapp/new-job/script.js
```diff
@@ -179,43 +209,109 @@ function getJobDetailsFromUserInput() {
...
+ * This function calculates and returns the annual pay depending
+ * on the maximum pay and the frequency.
+ * @param {int} max the upper limit on the pay.
+ * @param {String} frequency how often the employee will be paid.
+ * @return {int} the annual pay.
+ */
+function calculateAnnualMax(max, frequency) {
+  let annualMax = max; // default
+
+  switch (frequency) {
```
yohanlaunay: FYI (up to you), why not returning directly instead of annualMax = xxx + break;

walk-in-interview/src/main/webapp/new-job/script.js
```diff
@@ -229,42 +325,69 @@ function validateRequiredUserInput() {
...
+    setErrorMessageAndField(/* msg */ name.placeholder,
+        /** includes default msg */ true, /** error element id */ 'title');
+    return false;
   }
 
   if (description.value === '') {
-      setErrorMessage(/* errorMessageElementId= */'error-message', /* msg= */ description.placeholder);
-      return false;
+    setErrorMessageAndField(/* msg */ description.placeholder,
+        /** includes default msg */ true,
```
yohanlaunay: personally i would have changed the order:
```
setErrorMessageAndField(fieldId, errorMessage, includesDefault=true) <-- using a default value for that last field.
```

walk-in-interview/src/main/webapp/new-job/script.js
```diff
@@ -229,42 +325,69 @@ function validateRequiredUserInput() {
...
   if (address.value === '') {
-      setErrorMessage(/* errorMessageElementId= */'error-message', /* msg= */ address.placeholder);
-      return false;
+    setErrorMessageAndField(/* msg */ address.placeholder,
+        /** includes default msg */ true, /** error element id */ 'address');
+    return false;
+  }
+
+  if (postalCode.value === '' ||
+    (parseInt(postalCode.value[0]) > JAVA_INTEGER_MAX_VALUE) ||
```
yohanlaunay: it's not obvious why postal code first 2 characters have to be digits. you should document that. Also you are copying the parsing of postal code with 0 and 82. I would recommend instead you try to parse postalCode and if the parsing fails then setErrorMessageAndField instead of re-validating the values here.

walk-in-interview/src/main/webapp/script.js
```diff
@@ -12,10 +12,19 @@ const CurrentLocale = 'en';
  * TODO(issue/22): figure out how to use dynamic imports
  */
 import {AppStrings} from './strings.en.js';
+import {getRequirementsList} from './common-functions.js';
 
 const STRINGS = AppStrings['homepage'];
 const JOBPAGE_PATH = '/new-job/index.html';
-const RESPONSE_ERROR = 'An error occured while getting the job listings';
+const RESPONSE_ERROR = 'An error occured while getting the job listings.';
```
yohanlaunay: Anything that can be shown to the user must be in the STRINGS file.

walk-in-interview/src/main/webapp/script.js
```diff
@@ -122,60 +111,96 @@ function renderSelectOptions(id) {
...
+    return false;
+  }
+
+  /** If the field has been filled out, we check that it's a positive int */
+  if (!Number.isNaN(maxLimitParam) &&
+  (maxLimitParam > JAVA_INTEGER_MAX_VALUE || maxLimitParam < 0)) {
+    setErrorMessage(/* msg */ STRINGS['filter-max-limit'],
+        /** includes default msg */ true);
+    document.getElementById('filter-max-limit')
+        .style.backgroundColor = 'rgb(255,0,0, 0.1)';
```
yohanlaunay: could you apply a class instead? add/remove class.

walk-in-interview/src/main/webapp/script.js
```diff
@@ -187,33 +212,104 @@ async function renderJobListings(sortBy, order, pageSize, pageIndex) {
...
+async function renderJobListings() {
+  if (!validateFilters()) {
+    return;
+  }
+
+  const sortingParam = document.getElementById('sort-by').value;
+
+  /** Note that this platform currently only sorts by Salary.*/
+  if (!sortingParam.includes(SALARY_PARAM)) {
+    console.log('error', 'this app currently only sorts by salary');
```
yohanlaunay: `TODO(issue/)` support other sorting functions

walk-in-interview/src/main/webapp/script.js
```diff
@@ -187,33 +212,104 @@ async function renderJobListings(sortBy, order, pageSize, pageIndex) {
...
+
+  if (Number.isNaN(maxLimitParam)) {
+    maxLimitParam = JAVA_INTEGER_MAX_VALUE;
+  }
+
+  const jobPageData = await getJobListings(regionParam, sortByParam,
+      minLimitParam, maxLimitParam, orderByParam,
+      DEFAULT_PAGE_SIZE, DEFAULT_PAGE_INDEX)
+      .catch((error) => {
+        console.log('error', error);
```
yohanlaunay: nit, for javascript console log i recommend you put some more context since you don't have stacktraces:

`console.log('Error fetching job listings', error);`

walk-in-interview/src/main/java/com/google/job/data/JobQuery.java
```diff
@@ -0,0 +1,88 @@
...
+        return this;
+    }
+
+    public JobQuery setOrder(Order order) {
+        this.order = order;
+        return this;
+    }
+
+    public JobQuery setPageSize(int pageSize) {
+        assert pageSize >= 1;
```
yohanlaunay: i wrote that quickly as pseudocode, you should use proper if+throw IllegalArgumentException :)

walk-in-interview/src/main/java/com/google/job/data/JobsDatabase.java
```diff
@@ -138,4 +146,54 @@
...
+        SingaporeRegion region = jobQuery.getRegion();
+        if (!region.equals(SingaporeRegion.ENTIRE)) {
+            query = query.whereEqualTo(REGION_FIELD, region.name());
+        }
+
+        // TODO(issue/34): add to the query to include pagination (using pageSize and pageIndex)
+
+        return ApiFutures.transform(
+            query.get(),
+            new ApiFunction<QuerySnapshot, JobPage>() {
```
yohanlaunay: i'm surprised you cannot use a lambda here, you should be able to write:
```
return ApiFutures.transform(
    query.get(),
   querySnapshot -> {
      [...]
       return new JobPage(...);
   }, directExecutor());
```

walk-in-interview/src/main/java/com/google/job/data/JobsDatabase.java
```diff
@@ -138,4 +146,54 @@
...
+                @NullableDecl
+                public JobPage apply(@NullableDecl QuerySnapshot querySnapshot) {
+                    List<QueryDocumentSnapshot> documents = querySnapshot.getDocuments();
```
yohanlaunay: it seems querySnapshot could be null?

walk-in-interview/src/main/java/com/google/job/servlets/JobsListingsServlet.java
```diff
@@ -0,0 +1,229 @@
...
+    }
+
+    /**
+     * Returns the job query object.
+     *
+     * @param request From the GET request.
+     * @return the job query object.
+     * @throws IllegalArgumentException if one of the params is invalid.
+     */
+    public static JobQuery parseJobQuery(HttpServletRequest request) throws IllegalArgumentException {
```
yohanlaunay, line 81: this tells me you are always sending all the filters up from the client and they all must have a value. If that's expected then great, otherwise your parseXXX(request) methods should return optional for min, max, region, etc. (and also you can update your JobQuery to also take on optional fields in setXXX so that you don't need ifs everywhere)

walk-in-interview/src/main/webapp/new-job/script.js
```diff
@@ -179,43 +202,104 @@ function getJobDetailsFromUserInput() {
...
+      digits === 81) {
+    return 'EAST';
+  } else if ((digits >= 72 && digits <= 73) ||
+      (digits >= 75 && digits <= 80)) {
+    return 'NORTH';
+  } else if ((digits >= 53 && digits <= 57) ||
+      digits === 82) {
+    return 'NORTH_EAST';
+  }
+  setErrorMessageAndField(/* errorFieldId= */ 'postal-code',
```
yohanlaunay: i think the error message should be set by whoever calls findRegion instead of findRegion doing that itself. findRegion is not aware of the html fields (and rightly so).

walk-in-interview/src/main/webapp/new-job/script.js
```diff
@@ -179,43 +202,104 @@ function getJobDetailsFromUserInput() {
...
+ * @param {String} postalCode The Singapore postal code.
+ * @return {String} The Singapore region.
+ */
+function findRegion(postalCode) {
+  /*
+   * We have made the assumption that if the user's
+   * postal code region is 01, they would have written
+   * 01xxx rather than 1xxx.
+   */
+  const digits = parseInt(postalCode.substring(0, 2));
```
yohanlaunay, line 243: should you check that postalCode is at least 2 digits long?

walk-in-interview/src/main/java/com/google/job/data/JobPage.java
```diff
@@ -0,0 +1,108 @@
+package com.google.job.data;
+
+import java.util.*;
+import org.apache.commons.lang3.Range;
+import com.google.common.collect.ImmutableList;
+
+/** Class that represents the details of a page to show jobs. */
+public final class JobPage {
+    private final List<Job> jobList;
```
SibingWu, line 9: just wonder: here the jobList is the job posts in one page, right? e.g. jobs to be displayed for 30th - 50th job

walk-in-interview/src/main/java/com/google/job/data/JobPayment.java
```diff
@@ -5,20 +5,22 @@
...
 
-    public JobPayment(int min, int max, PaymentFrequency paymentFrequency) {
+    public JobPayment(int min, int max, PaymentFrequency paymentFrequency, long annualMax) {
         validateParameter(min, max);
 
         this.min = min;
         this.max = max;
         this.paymentFrequency = paymentFrequency;
+        this.annualMax = annualMax;
     }
```
SibingWu: I suppose it would be better just pass in (min, max, paymentFrequency) and count `annualMax` in the constructor according to `max` and `paymentFrequency`?

walk-in-interview/src/main/java/com/google/job/data/Order.java
```diff
@@ -0,0 +1,46 @@
...
+        for (Order order: values()){
+            if (order.getOrderId().equals(id)){
+                return order;
+            }
+        }
+
+        throw new IllegalArgumentException("Invalid order id: " + id);
+    }
+
+    public static Direction getQueryDirection(Order order) {
```
SibingWu, line 37: nit: javadoc

walk-in-interview/src/main/java/com/google/job/servlets/JobServlet.java
```diff
@@ -16,7 +16,10 @@
 import java.util.concurrent.TimeoutException;
 import java.util.stream.Collectors;
 
-/** Servlet that handles posting new job posts and updating existing job posts. */
+/**
+ * Servlet that handles posting new job posts, updating existing job posts,
+ * and getting an individual job listing.
```
SibingWu: nit: getting an individual job `post`

walk-in-interview/src/main/java/com/google/job/servlets/JobsListingsServlet.java
```diff
@@ -0,0 +1,229 @@
...
+public final class JobsListingsServlet extends HttpServlet {
+    private static final long TIMEOUT = 5;
+
+    public static final String MIN_LIMIT_PARAM = "minLimit";
+    public static final String MAX_LIMIT_PARAM = "maxLimit";
+    public static final String REGION_PARAM = "region";
+    public static final String SORT_BY_PARAM = "sortBy";
+    public static final String ORDER_PARAM = "order";
+    public static final String PAGE_SIZE_PARAM = "pageSize";
+    public static final String PAGE_INDEX_PARAM = "pageIndex";
```
SibingWu: why these need to be public? seems i did not find other places except this servlet using these constants?

walk-in-interview/src/main/java/com/google/job/servlets/JobsListingsServlet.java
```diff
@@ -0,0 +1,229 @@
...
+
+    @Override
+    public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+        try {
+            
+            JobQuery jobQuery = parseJobQuery(request);
+
+            JobPage jobPage = fetchJobPageDetails(jobQuery);
+
+            String json = new Gson().toJson(jobPage);
```
SibingWu: nit: there is a `convertToJsonUsingGson` in `ServletUtils.java`

walk-in-interview/src/main/java/com/google/job/servlets/JobsListingsServlet.java
```diff
@@ -0,0 +1,229 @@
...
+
+    /**
+     * Returns the lower limit as an int.
+     *
+     * @param request From the GET request.
+     * @return the lower limit.
+     * @throws IllegalArgumentException if the page size is invalid.
+     */
+    public static int parseMinLimit(HttpServletRequest request) throws IllegalArgumentException {
+        String minLimitStr = (String) request.getParameter(MIN_LIMIT_PARAM);
```
SibingWu: nit: there is a `getStringParameter` in `ServletUtils.java`
same for below

walk-in-interview/src/main/webapp/script.js
```diff
@@ -122,98 +109,205 @@ function renderSelectOptions(id) {
...
+    (minLimitParam > JAVA_INTEGER_MAX_VALUE || minLimitParam < 0)) {
+    setErrorMessage(/* msg= */ STRINGS['filter-min-limit'],
+        /* includesDefaultMsg= */ true);
+    document.getElementById('filter-min-limit').classList.add('error-field');
+    return false;
+  }
+
+  /* If the field has been filled out, we check that it's a positive int */
+  if (!Number.isNaN(maxLimitParam) &&
+  (maxLimitParam > JAVA_INTEGER_MAX_VALUE || maxLimitParam < 0)) {
```
SibingWu: nit: here seems need indentation

walk-in-interview/src/main/java/com/google/job/data/JobPage.java
riya-n: yep, I haven't implemented pagination as yet though

walk-in-interview/src/main/java/com/google/job/data/JobPayment.java
riya-n: I tried doing that, but for some reason it doesn't get properly stored in the database.

walk-in-interview/src/main/java/com/google/job/data/JobsDatabase.java
riya-n: yep, I keep getting errors when I try it like that

walk-in-interview/src/main/java/com/google/job/servlets/JobsListingsServlet.java
riya-n: yep the request will always have all filters with values so no need for optionals :)

yohanlaunay: noted. thanks for clarifying.

walk-in-interview/src/main/java/com/google/job/servlets/JobsListingsServlet.java
```diff
@@ -0,0 +1,229 @@
...
+     * @throws IllegalArgumentException if the id is invalid.
+     */
+    public static Order parseOrder(HttpServletRequest request) throws IllegalArgumentException {
+        String orderId = ServletUtils.getStringParameter(request, ORDER_PARAM, /* defaultValue= */ "");
+
+        if (orderId.isEmpty()) {
+            throw new IllegalArgumentException("order param should not be empty");
+        }
+
+        return Order.getFromId(orderId); // Illegal Argument Exception may be thrown
```
yohanlaunay: nit: `// IllegalArgumentException may be thrown`

walk-in-interview/src/main/webapp/new-job/script.js
```diff
@@ -179,43 +202,103 @@ function getJobDetailsFromUserInput() {
...
+    return 'EAST';
+  } else if ((digits >= 72 && digits <= 73) ||
+      (digits >= 75 && digits <= 80)) {
+    return 'NORTH';
+  } else if ((digits >= 53 && digits <= 57) ||
+      digits === 82) {
+    return 'NORTH_EAST';
+  }
+
+  throw new Error('invalid postal code');
```
yohanlaunay: nit, i recommend you also add the postal code to the error: `throw new Error('invalid postal code:'+postalCode);`

walk-in-interview/src/test/java/com/google/job/data/JobsDatabaseTest.java
```diff
@@ -245,6 +246,182 @@ public void fetchJob_normalInput_success() throws ExecutionException, Interrupte
         assertEquals(job, actualJob);
     }
 
+    @Test
+    public void fetchJobPage_normalInput_success() throws ExecutionException, InterruptedException, IOException {
+        /* fields that don't affect the job page details */
```
yohanlaunay: nit: `// Arrange`

walk-in-interview/src/test/java/com/google/job/data/JobsDatabaseTest.java
```diff
@@ -245,6 +246,182 @@ public void fetchJob_normalInput_success() throws ExecutionException, Interrupte
...
+        JobPage expectedJobPage = new JobPage(/* jobList= */ Arrays.asList(job1, job2, job3),
+            /* totalCount= */ 3, Range.between(1, 3));
+
+        // sorting is already defaulted to SALARY and ordering is defaulted to DESCENDING
+        // maxLimit is defaulted to Integer.MAX_VALUE
+        JobQuery jobQuery = new JobQuery().setMinLimit(104001).setRegion(SingaporeRegion.CENTRAL);
+        
+        JobPage actualJobPage = jobsDatabase.fetchJobPage(jobQuery).get();
+
+        assertEquals(expectedJobPage, actualJobPage);
```
yohanlaunay, line 332: nit: `// Assert`

walk-in-interview/src/test/java/com/google/job/data/JobsDatabaseTest.java
```diff
@@ -245,6 +246,182 @@ public void fetchJob_normalInput_success() throws ExecutionException, Interrupte
...
+    }
+
+     @Test
+    public void fetchJobPage_noJobsFitFilters_success() throws ExecutionException, InterruptedException, IOException {
+        /* fields that don't affect the job page details */
```
yohanlaunay: `// Arrange`

walk-in-interview/src/test/java/com/google/job/data/JobsDatabaseTest.java
```diff
@@ -245,6 +246,182 @@ public void fetchJob_normalInput_success() throws ExecutionException, Interrupte
...
+                .setJobPay(jobPayment2)
+                .build();
+    
+        firestore.collection(TEST_JOB_COLLECTION).add(job1).get();
+        firestore.collection(TEST_JOB_COLLECTION).add(job2).get();
+
+        JobPage expectedJobPage = new JobPage(/* jobList= */ Arrays.asList(),
+            /* totalCount= */ 0, Range.between(0, 0));
+
+        // sorting is already defaulted to SALARY and ordering is defaulted to DESCENDING
```
yohanlaunay, line 364: `// Act.`

walk-in-interview/src/test/java/com/google/job/data/JobsDatabaseTest.java
```diff
@@ -245,6 +246,182 @@ public void fetchJob_normalInput_success() throws ExecutionException, Interrupte
...
+        // minLimit is defaulted to 0 and maxLimit is defaulted to Integer.MAX_VALUE
+        JobQuery jobQuery = new JobQuery().setRegion(SingaporeRegion.NORTH);
+        
+        JobPage actualJobPage = jobsDatabase.fetchJobPage(jobQuery).get();
+
+        assertEquals(expectedJobPage, actualJobPage);
```
yohanlaunay, line 371: `// Assert.`

walk-in-interview/src/test/java/com/google/job/data/JobsDatabaseTest.java
```diff
@@ -245,6 +246,182 @@ public void fetchJob_normalInput_success() throws ExecutionException, Interrupte
...
+        firestore.collection(TEST_JOB_COLLECTION).add(job1).get();
+        firestore.collection(TEST_JOB_COLLECTION).add(job3).get();
+        firestore.collection(TEST_JOB_COLLECTION).add(job4).get();
+        firestore.collection(TEST_JOB_COLLECTION).add(job2).get();
+        firestore.collection(TEST_JOB_COLLECTION).add(job6).get();
+
+        JobPage expectedJobPage = new JobPage(/* jobList= */ Arrays.asList(job1, job2, job3),
+            /* totalCount= */ 3, Range.between(1, 3));
+
+        // sorting is already defaulted to SALARY and ordering is defaulted to DESCENDING
```
yohanlaunay, line 325: nit: `// Act`
^ specially because your Arrange section is really big (probably too big not sure if you can simplify a bit)

walk-in-interview/src/test/java/com/google/job/data/JobsDatabaseTest.java
```diff
@@ -245,6 +246,182 @@ public void fetchJob_normalInput_success() throws ExecutionException, Interrupte
...
+        // maxLimit is defaulted to Integer.MAX_VALUE
+        JobQuery jobQuery = new JobQuery().setMinLimit(104001).setRegion(SingaporeRegion.CENTRAL);
```
yohanlaunay: 104001 <-- i recommend you define a variable and use that variable in the query and in the job definition or something like that. It can be the exact job salary or it can be something like `int minLimit = job1Salary - 1;`. However you prefer, the idea being that you want to link that value to the data that you arranged above.

yohanlaunay (changes requested):

SibingWu (approved):

walk-in-interview/src/main/java/com/google/job/data/JobPage.java
```diff
@@ -0,0 +1,108 @@
...
+
+    @Override
+    public String toString() {
+        return String.format("JobPage{jobList=%s, totalCount=%d, range=%s}",
+                jobList, totalCount, range);
+    }
+
+    private static void validateParameters(List<Job> jobList, long totalCount, Range<Integer> range)
+            throws IllegalArgumentException {
+        if (jobList.size() < 0) {
```
htthach: Will this ever happen?

walk-in-interview/src/main/java/com/google/job/data/JobPage.java
```diff
@@ -0,0 +1,108 @@
...
+        }
+
+        if (totalCount < 0 || totalCount < jobList.size()) {
+            throw new IllegalArgumentException("totalCount should not be negative or less than jobList size");
+        }
+
+        if (range.getMinimum() < 0 || range.getMaximum() > totalCount ||
+                range.getMaximum() < range.getMinimum()) {
+            throw new IllegalArgumentException("range should not have invalid max/min values");
+        }
```
htthach: How about the relationship between range max, min vs jobList.size ?

walk-in-interview/src/main/java/com/google/job/data/JobPayment.java
```diff
@@ -5,20 +5,22 @@
...
 
-    public JobPayment(int min, int max, PaymentFrequency paymentFrequency) {
+    public JobPayment(int min, int max, PaymentFrequency paymentFrequency, long annualMax) {
         validateParameter(min, max);
 
         this.min = min;
         this.max = max;
         this.paymentFrequency = paymentFrequency;
+        this.annualMax = annualMax;
```
htthach: Currently validateParameter doesn't validate if the annualMax consistent with the previous max * frequency.

walk-in-interview/src/main/java/com/google/job/data/JobQuery.java
```diff
@@ -0,0 +1,100 @@
+package com.google.job.data;
+
+/** Class for the job listings query using builder pattern. */
+public final class JobQuery {
```
htthach, line 4: This class look like a builder, it's supposed to build some immutable object, are we missing that part?

walk-in-interview/src/main/java/com/google/job/data/Location.java
```diff
@@ -5,18 +5,20 @@
...
+    private final SingaporeRegion region;
     private final double latitude;
     private final double longitude;
 
     private volatile int hashCode;
 
     // For serialization
     public Location() {
-        this(/* address= */"", /* postalCode= */"dummy",
+        this(/* address= */"", /* postalCode= */"dummy", /* region= */ SingaporeRegion.CENTRAL,
```
htthach: Is central the default or entire?

walk-in-interview/src/main/java/com/google/job/servlets/JobsListingsServlet.java
```diff
@@ -0,0 +1,229 @@
...
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.*;
+
+/**
+ * Servlet that handles getting the job listings given the filters.
+ */
+@WebServlet("/jobs/listings")
+public final class JobsListingsServlet extends HttpServlet {
+    private static final long TIMEOUT = 5;
```
htthach: Nit: TIMEOUT_SECONDS?

walk-in-interview/src/main/java/com/google/job/servlets/JobsListingsServlet.java
```diff
@@ -0,0 +1,229 @@
...
+    }
+
+    /**
+     * Returns the region as a SingaporeRegion enum.
+     *
+     * @param request From the GET request.
+     * @return SingaporeRegion enum given the id in the request params.
+     * @throws IllegalArgumentException if the id is invalid.
+     */
+    private static SingaporeRegion parseRegion(HttpServletRequest request) throws IllegalArgumentException {
```
htthach: If no region, we assume entire instead of throw?

walk-in-interview/src/main/java/com/google/job/servlets/JobsListingsServlet.java
```diff
@@ -0,0 +1,229 @@
...
+
+    /**
+     * Returns the lower limit as an int.
+     *
+     * @param request From the GET request.
+     * @return the lower limit.
+     * @throws IllegalArgumentException if the page size is invalid.
+     */
+    public static int parseMinLimit(HttpServletRequest request) throws IllegalArgumentException {
+        String minLimitStr = ServletUtils.getStringParameter(request, MIN_LIMIT_PARAM, /* defaultValue= */ "");
```
htthach, line 105: Would the query work without min and max? e.g if not supplied, don't apply those filters? Look like we have some default value set for query builder above?

walk-in-interview/src/main/java/com/google/job/servlets/JobsListingsServlet.java
```diff
@@ -0,0 +1,229 @@
...
+
+    /**
+     * Returns the page size as an int.
+     *
+     * @param request From the GET request.
+     * @return the page size.
+     * @throws IllegalArgumentException if the page size is invalid.
+     */
+    public static int parsePageSize(HttpServletRequest request) throws IllegalArgumentException {
+        String pageSizeStr = ServletUtils.getStringParameter(request, PAGE_SIZE_PARAM, /* defaultValue= */ "");
```
htthach, line 197: Would it be better to default to some value instead of throw? Or are we currently trying to be explicit in all parameters?

walk-in-interview/src/main/java/com/google/job/data/JobQuery.java
```diff
@@ -0,0 +1,100 @@
...
+    private int minLimit = 0;
```
htthach, line 5: How about unset value? E.g if I build a query and did not set the min or max and you don't apply those condition? Is it supported? Or are we requiring all query conditions to be set or else fallback to these default initialized value?

walk-in-interview/src/main/webapp/new-job/script.js
```diff
@@ -228,43 +311,85 @@ function validateRequiredUserInput() {
...
+  }
+
+  /*
+   * The first two digits of the postal code must be numbers within
+   * the below range because those two digits correspond to the location's
+   * district, which indicates its region in Singapore.
+   */
+  if (postalCode.value === '' || postalCode.length < 2 ||
+    (parseInt(postalCode.value[0]) > JAVA_INTEGER_MAX_VALUE) ||
+    (parseInt(postalCode.value[1]) > JAVA_INTEGER_MAX_VALUE)) {
```
htthach: How can a digit be more than Max Int?

walk-in-interview/src/main/webapp/script.js
```diff
@@ -122,98 +109,205 @@ function renderSelectOptions(id) {
...
+
+  if (Number.isNaN(minLimitParam)) {
+    minLimitParam = 0;
+  }
+
+  if (Number.isNaN(maxLimitParam)) {
+    maxLimitParam = JAVA_INTEGER_MAX_VALUE;
+  }
+
+  const jobPageData = await getJobListings(regionParam, sortByParam,
```
htthach, line 281: The function said async but here we do await which is sync?

walk-in-interview/src/main/java/com/google/job/data/JobPage.java
riya-n: I think the .size() function can be overriden right?

walk-in-interview/src/main/java/com/google/job/data/Location.java
riya-n: entire

walk-in-interview/src/main/webapp/script.js
riya-n: I thought the async/await keywords were used to make the code act more synchronously?

walk-in-interview/src/main/java/com/google/job/data/JobQuery.java
riya-n: If some query conditions aren't set, then they will be defaulted to these values

walk-in-interview/src/main/java/com/google/job/data/JobQuery.java
riya-n: I think we're just using the builder pattern here rather than an explicit builder.

walk-in-interview/src/main/java/com/google/job/servlets/JobsListingsServlet.java
riya-n: Yep, it would still work without min/max because it would default to the JobQuery value. But I thought it might make sense to have all the filters in the request

walk-in-interview/src/main/java/com/google/job/servlets/JobsListingsServlet.java
riya-n: I think I was trying to be explicit with all the parameters. I will set the region to ENTIRE is there is no region though.

riya-n: Currently pageSize and pageIndex are the default values though. Since pagination hasn't been implemented yet

walk-in-interview/src/main/java/com/google/job/servlets/JobsListingsServlet.java
riya-n: If the min/max aren't supplied, then they should just become the default values - 0 and Integer.MAX_VALUE

walk-in-interview/src/main/webapp/script.js
```diff
@@ -122,98 +109,205 @@ function renderSelectOptions(id) {
...
+  }
 
-    const sortByParam = document.getElementById('sort-by').value;
-    const sortOrderParam = document.getElementById('sort-by-order')
-        .value;
-    renderJobListings(sortByParam, sortOrderParam, DEFAULT_PAGE_SIZE,
-        DEFAULT_PAGE_INDEX);
-  });
+  /* If the field has been filled out, we check that it's a positive int */
+  if (!Number.isNaN(minLimitParam) &&
```
htthach, line 130: What if either limit is not a number?

walk-in-interview/src/main/webapp/script.js
```diff
@@ -122,98 +109,205 @@ function renderSelectOptions(id) {
...
+        /* includesDefaultMsg= */ false);
+    return;
+  }
+
+  const jobListings = jobPageData['jobList'];
+
   const jobListingTemplate = document.getElementById('job-listing-template');
 
   jobListings.forEach((job) => {
     const jobListing =
```
htthach: Should we abstract the body of this block to another method like 
forEach((job) => { jobListingsElement.appendChild(buildJobElement(job))})

walk-in-interview/src/test/java/com/google/job/data/JobsDatabaseTest.java
```diff
@@ -245,6 +246,189 @@ public void fetchJob_normalInput_success() throws ExecutionException, Interrupte
...
+        // Arrange
+        /* fields that don't affect the job page details */
+        JobStatus jobStatus = JobStatus.ACTIVE;
+        String jobName = "Programmer";
+        String jobDescription = "Fighting to defeat hair line recede";
+        List<String> requirements = Requirement.getLocalizedNames(Arrays.asList(O_LEVEL), "en");
+        long postExpiry = System.currentTimeMillis();
+        JobDuration jobDuration = JobDuration.ONE_MONTH;
+
+        /* fields that affect the sorting/filtering of the job page details */
```
htthach, line 253: Can we abstract these data setup code into another method? 
e.g List<Job> testJobs = createTestJobs(/* count= */ 5);

walk-in-interview/src/test/java/com/google/job/data/JobsDatabaseTest.java
```diff
@@ -245,6 +246,189 @@ public void fetchJob_normalInput_success() throws ExecutionException, Interrupte
...
+        Location location1 =  new Location("Maple Tree", "123456", SingaporeRegion.CENTRAL, 0, 0);
```
htthach, line 339: If we do the abstraction above, here we can do List<Job> testJobs = createTestJobs(/*count=*/ 2) ?

walk-in-interview/src/main/webapp/new-job/script.js
```diff
@@ -228,43 +311,85 @@ function validateRequiredUserInput() {
...
+        /* msg= */ address.placeholder, /* includesDefaultMsg= */ true);
+    return false;
+  }
+
+  /*
+   * The first two digits of the postal code must be numbers within
+   * the below range because those two digits correspond to the location's
+   * district, which indicates its region in Singapore.
+   */
+  if (postalCode.value === '' || postalCode.length < 2 ||
```
htthach: Since postal code validation is not trivial, should we abstract it out to another method?

walk-in-interview/src/main/webapp/script.js
```diff
@@ -122,98 +109,205 @@ function renderSelectOptions(id) {
...
+  const sortByParam = SALARY_PARAM;
+  const orderByParam = sortingParam.substring(7);
+
+  const regionParam = document.getElementById('show-by-region').value;
+  let minLimitParam = document.getElementById('filter-min-limit')
+      .valueAsNumber;
+  let maxLimitParam = document.getElementById('filter-max-limit')
+      .valueAsNumber;
+
+  if (Number.isNaN(minLimitParam)) {
```
htthach, line 273: Since we do this check here instead of inside validateFilters. Does that means that we allow NaN in the limit? Can user enter by mistake '100sgd'? And if we fallback to default values when they're NaN, do we reset those values to the text field?

walk-in-interview/src/main/webapp/script.js
```diff
@@ -122,98 +109,205 @@ function renderSelectOptions(id) {
...
+  jobShowing.innerText = `${jobPageData['range'].minimum} -` +
+    ` ${jobPageData['range'].maximum} ${STRINGS['job-listings-showing']} ` +
+    `${jobPageData['totalCount']}`;
+}
+
+/**
+ * Add the list of jobs that are stored in the database
+ * given the filter fields.
+ */
+async function renderJobListings() {
```
htthach: Nit: this method look like it's loadAndDisplayJobListing?

walk-in-interview/src/main/webapp/script.js
```diff
@@ -122,98 +109,205 @@ function renderSelectOptions(id) {
...
+  /*
+   * Note that this platform currently only sorts by Salary.
+   * TODO(issue/62): support more filters
+   */
+  if (!sortingParam.includes(SALARY_PARAM)) {
+    console.error('this app currently only sorts by salary');
+    return;
+  }
+
+  const sortByParam = SALARY_PARAM;
```
htthach, line 270: Maybe we can get these param out from the document element here, then pass these param to the validateFilterParams(param1, param2...) 
then call
getJobListing(param1, param2...)
This way we're sure that everyone is looking at the same param?

walk-in-interview/src/main/java/com/google/job/data/JobPage.java
yohanlaunay: this could happen. i recommend isEmpty() instead though (super nit)

walk-in-interview/src/main/java/com/google/job/data/JobPage.java
```diff
@@ -0,0 +1,116 @@
...
+
+        if (jobList.size() != 0 && (range.getMaximum() - range.getMinimum() + 1) != jobList.size()) {
```
yohanlaunay: you already checked for size above

walk-in-interview/src/main/java/com/google/job/data/JobPage.java
```diff
@@ -0,0 +1,116 @@
...
+            throw new IllegalArgumentException("range should reflect jobList size");
+        }
+
+        if (jobList.size() == 0 && (range.getMaximum() != 0 || range.getMinimum() != 0)) {
```
yohanlaunay: you already checked for size above, this condition is not possible

walk-in-interview/src/main/java/com/google/job/data/JobPayment.java
```diff
@@ -5,20 +5,27 @@
...
 
     // For serialization
     public JobPayment() {
-        this(/* min= */0, /* max= */0, PaymentFrequency.HOURLY);
+        this(/* min= */0, /* max= */0, PaymentFrequency.HOURLY, /* annualMax= */ 0);
     }
 
-    public JobPayment(int min, int max, PaymentFrequency paymentFrequency) {
-        validateParameter(min, max);
+    public JobPayment(int min, int max, PaymentFrequency paymentFrequency, long annualMax) {
```
yohanlaunay: FYI, after this param you should not add any more params to this class. if you need to add another one, you should change to a Builder pattern JobPayment.Builder

walk-in-interview/src/main/java/com/google/job/data/JobPayment.java
```diff
@@ -61,24 +74,50 @@ public int hashCode() {
...
+                break;
+            case MONTHLY:
+                expectedAnnualMax = max * MONTHS_PER_YEAR;
+                break;
+            case YEARLY:
+                expectedAnnualMax = max;
+                break;
+        }
+
+        if (annualMax != expectedAnnualMax) {
```
yohanlaunay: if you are able to compute annualMax why are you passing it in?

walk-in-interview/src/main/java/com/google/job/data/JobPayment.java
```diff
@@ -61,24 +74,50 @@ public int hashCode() {
...
     @Override
     public String toString() {
-        return String.format("JobPayment{min=%d, max=%d, paymentFrequency=%s}",
-                min, max, paymentFrequency.name());
+        return String.format("JobPayment{min=%d, max=%d, paymentFrequency=%s, annualMax=%d}",
+                min, max, paymentFrequency.name(), annualMax);
     }
 
-    private static void validateParameter(float min, float max) throws IllegalArgumentException {
+    private static void validateParameter(float min, float max, PaymentFrequency paymentFrequency, long annualMax)
```
yohanlaunay: the more i read those validateParameter methods you have everywhere the more i think it should be inlined in the constructor. separating the param validation from constructor forces you to pass all the arguments(so the method gets longer and longer) when it would be more readable to put this in the constructor directly at least you would know right there was is a valid param instead of deferring the knowledge and forcing someone to find that out by scrollin gall the way down. And if any validation is complicated and requires pre computation (beyond an if+throw) then the computation needed for the validation should be moved to a method. but i would recommend you inline this code instead.

walk-in-interview/src/main/java/com/google/job/data/JobsDatabase.java
```diff
@@ -138,4 +146,60 @@
...
+            query = query.whereEqualTo(REGION_FIELD, region.name());
+        }
+
+        // TODO(issue/34): add to the query to include pagination (using pageSize and pageIndex)
+
+        return ApiFutures.transform(
+            query.get(),
+            querySnapshot -> {
+                if (querySnapshot == null) {
+                    return new JobPage(ImmutableList.of(), 0, Range.between(0, 0));
```
yohanlaunay: please annotate empty lists and naked ints. (same below) `/* paramName= */ImmutableList.of()`

walk-in-interview/src/main/java/com/google/job/servlets/JobsListingsServlet.java
```diff
@@ -0,0 +1,228 @@
...
+     */
+    public static JobQuery parseJobQuery(HttpServletRequest request) throws IllegalArgumentException {
+        int minLimit = parseMinLimit(request);
+        int maxLimit = parseMaxLimit(request);
+        SingaporeRegion region = parseRegion(request);
+        Filter sortBy = parseSortBy(request);
+        Order order = parseOrder(request);
+        // TODO(issue/34): parse page size and index once pagination is implemented
+
+        JobQuery jobQuery = new JobQuery().setMinLimit(minLimit).setMaxLimit(maxLimit).setRegion(region)
```
yohanlaunay: nit: i feel this should have been formatted. Typically it would look like this (when formatted automatically):
```
JobQuery jobQuery = new JobQuery()
                                           .setMinLimit(minLimit)
                                           .setMaxLimit(minLimit)
[...]
```

also you can use return directly:
```
return new JobQuery()
                   .setMinLimit(minLimit)
                    .setMaxLimit(minLimit)
[...]
```

walk-in-interview/src/main/webapp/script.js
riya-n: So the user can only type in numbers. If they choose to leave the limit blank, then the default values will get sent to the servlet.

walk-in-interview/src/main/java/com/google/job/data/JobPayment.java
riya-n: noted, thanks

walk-in-interview/src/main/webapp/script.js
riya-n: The user should only be able to type numbers into the text field (the input type is number), but even if they type something else, then the minLimit and maxLimit will just reset to the default values.

riya-n: The text fields are just left as blank for the default values. I thought this would be better since it doesn't make sense to explicitly write min=0 and max=214...(whatever the max integer value is for java). Do you think it would be better to have the values filled in?

walk-in-interview/src/main/java/com/google/job/data/JobPage.java
riya-n: when there are no jobs that match the filters, the JobPage object to be returned should have all these fields as 0.

walk-in-interview/src/main/java/com/google/job/data/JobPage.java
riya-n: this is to check the relationship between jobList size and range min/max (the relationship is different if the joblist size is 0)

walk-in-interview/src/main/java/com/google/job/data/JobPayment.java
```diff
@@ -5,20 +5,39 @@
...
+        long expectedAnnualMax = findExpectedAnnualMax(max, paymentFrequency);
+
+        if (annualMax != expectedAnnualMax) {
+            throw new IllegalArgumentException("annualMax was : " + annualMax + ", but should be: " + expectedAnnualMax);
+        }
 
         this.min = min;
         this.max = max;
         this.paymentFrequency = paymentFrequency;
+        this.annualMax = annualMax;
```
yohanlaunay: i recommend: `this.annualMax = findExpectedAnnualMax(max, paymentFrequency);` instead of using a param. this is essentially equivalent since you are throwing an IllegalArg if the param is not the same as your computation.

walk-in-interview/src/main/webapp/new-job/script.js
yohanlaunay: i still don't see the check. unresolving :p

yohanlaunay (changes requested):

yohanlaunay (approved):

PR #64, opened 2020-07-21
Add Requirement filter

### LINKED ISSUE #65: Add requirement filter
Add a requirement filter so that an applicant without driving license will not see the job that requires a driving license

SibingWu:
Fixed #65 

* Add a requirement filter so that an applicant without driving license will not see the job that requires a driving license
* Change the requirement schema for job post from list to `requirementTag: true/false`
    * Limitation: Assumes that the requirement list will not extend. If there needs to be more requirements, data migration is needed to update all the existing job posts with new requirement fields. E.g. there is a new requirement naming “HEIGHT”, then all existing job post `requirements` field need to update and add “HEIGHT”: false

* Write tests for `fetchAllEligibleJobs`

### DIFF
--- a/walk-in-interview/src/main/java/com/google/job/data/Job.java
+++ b/walk-in-interview/src/main/java/com/google/job/data/Job.java
@@ -70,7 +70,7 @@ public final class Job {
     public static final class JobBuilder {
         // Optional parameters - initialized to default values
         private String jobId = "";
-        private Map<String, Boolean> requirements = ImmutableMap.of(); // Maps with non-string keys are not supported
+        private Map<String, Boolean> requirements = ImmutableMap.of();
         private JobDuration jobDuration = JobDuration.OTHER;
 
         // TODO(issue/25): merge the account stuff into job post.
@@ -203,7 +203,7 @@ public final class Job {
         return jobPay;
     }
 
-    /** Returns a list of requirements (stable ids) of the job post. */
+    /** Returns a map of requirements (stable ids) of the job post. */
     public Map<String, Boolean> getRequirements() {
         return requirements;
     }
--- a/walk-in-interview/src/main/java/com/google/job/data/Job.java
+++ b/walk-in-interview/src/main/java/com/google/job/data/Job.java
@@ -14,7 +14,7 @@ public final class Job {
     private final Location jobLocation;
     private final String jobDescription;
     private final JobPayment jobPay;
-    private final Map<Requirement, Boolean> requirements; // requirement enum value : true/false
+    private final Map<String, Boolean> requirements; // requirement stable id : true/false
     private final long postExpiryTimestamp;
     private final JobDuration jobDuration;
 
@@ -70,7 +70,7 @@ public final class Job {
     public static final class JobBuilder {
         // Optional parameters - initialized to default values
         private String jobId = "";
-        private Map<Requirement, Boolean> requirements = ImmutableMap.of();
+        private Map<String, Boolean> requirements = ImmutableMap.of(); // Maps with non-string keys are not supported
         private JobDuration jobDuration = JobDuration.OTHER;
 
         // TODO(issue/25): merge the account stuff into job post.
@@ -129,16 +129,16 @@ public final class Job {
             return this;
         }
 
-        public JobBuilder setRequirements(List<Requirement> requirements) {
+        public JobBuilder setRequirements(List<String> requirements) {
             // Changes requirements into set to make "contains" operation O(1)
-            Set<Requirement> requirementsSet = new HashSet<>(requirements);
+            Set<String> requirementsSet = new HashSet<>(requirements);
 
-            // Gets all requirements
-            List<Requirement> requirementsList = Arrays.asList(Requirement.values());
+            // Gets all requirements stable id
+            List<String> requirementsList = Requirement.getAllRequirementIds();
 
-            ImmutableMap.Builder<Requirement, Boolean> requirementsMap = ImmutableMap.builder();
+            ImmutableMap.Builder<String, Boolean> requirementsMap = ImmutableMap.builder();
 
-            for (Requirement requirement: requirementsList) {
+            for (String requirement: requirementsList) {
                 requirementsMap.put(requirement, requirementsSet.contains(requirement));
             }
 
@@ -216,8 +216,8 @@ public final class Job {
         return jobPay;
     }
 
-    /** Returns a list of requirements of the job post. */
-    public Map<Requirement, Boolean> getRequirements() {
+    /** Returns a list of requirements (stable ids) of the job post. */
+    public Map<String, Boolean> getRequirements() {
         return requirements;
     }
 
--- a/walk-in-interview/src/main/java/com/google/job/data/Job.java
+++ b/walk-in-interview/src/main/java/com/google/job/data/Job.java
@@ -129,21 +129,8 @@ public final class Job {
             return this;
         }
 
-        public JobBuilder setRequirements(List<String> requirements) {
-            // Changes requirements into set to make "contains" operation O(1)
-            Set<String> requirementsSet = new HashSet<>(requirements);
-
-            // Gets all requirements stable id
-            List<String> requirementsList = Requirement.getAllRequirementIds();
-
-            ImmutableMap.Builder<String, Boolean> requirementsMap = ImmutableMap.builder();
-
-            for (String requirement: requirementsList) {
-                requirementsMap.put(requirement, requirementsSet.contains(requirement));
-            }
-
-            this.requirements = requirementsMap.build();
-
+        public JobBuilder setRequirements(Map<String, Boolean> requirements) {
+            this.requirements = ImmutableMap.copyOf(requirements);
             return this;
         }
 
--- a/walk-in-interview/src/main/java/com/google/job/data/Job.java
+++ b/walk-in-interview/src/main/java/com/google/job/data/Job.java
@@ -130,53 +130,23 @@ public final class Job {
         }
 
         public JobBuilder setRequirements(Map<String, Boolean> requirements) {
-            List<String> requirementsList = Requirement.getAllRequirementIds();
+            Map<String, Boolean> inputRequirementsMap = new HashMap<>(requirements);
+            Set<String> requirementsSet = new HashSet<>(Requirement.getAllRequirementIds());
 
-            if (requirements.size() == requirementsList.size()) {
-                this.requirements = ImmutableMap.copyOf(requirements);
-                return this;
-            } else if (requirements.size() < requirements.size()) {
-                // some requirements are not contained when sent from client
+            // Removes redundant requirement
+            inputRequirementsMap.entrySet().removeIf(entry -> !requirementsSet.contains(entry.getKey()));
 
-                Map<String, Boolean> requirementsMap = new HashMap<>(requirements);
-
-                for (String requirement: requirementsList) {
-                    if (requirementsMap.containsKey(requirement)) {
-                        continue;
-                    }
-
-                    requirementsMap.put(requirement, false);
+            // Checks if requirement missing and adds accordingly
+            for (String requirement: requirementsSet) {
+                if (inputRequirementsMap.containsKey(requirement)) {
+                    continue;
                 }
 
-                this.requirements = ImmutableMap.copyOf(requirementsMap);
-                return this;
-            } else {
-                // excessive requirements are sent from the client
-
-                Map<String, Boolean> requirementsMap = new HashMap<>(requirements);
-                Set<String> requirementsSet = new HashSet<>(requirementsList);
-
-                // Removes redundant requirement
-                for (String requirement: requirementsMap.keySet()) {
-                    if (requirementsSet.contains(requirement)) {
-                        continue;
-                    }
-
-                    requirementsMap.remove(requirement);
-                }
-
-                // Checks if requirement missing and adds accordingly
-                for (String requirement: requirementsSet) {
-                    if (requirementsMap.containsKey(requirement)) {
-                        continue;
-                    }
-
-                    requirementsMap.put(requirement, false);
-                }
-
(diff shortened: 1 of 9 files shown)


walk-in-interview/src/main/java/com/google/job/data/JobsDatabase.java
```diff
@@ -138,4 +140,57 @@
...
+
+        ApiFunction<QuerySnapshot, Collection<Job>> function = documents -> {
+            ImmutableSet.Builder<Job> jobs = ImmutableSet.builder();
+
+            for (DocumentSnapshot document : documents) {
+                Job job = document.toObject(Job.class);
+                List<String> requirements = job.getRequirements();
+
+                // Eligible job: all job requirements are contained in the applicant skills.
+                if (skills.containsAll(requirements)) {
```
riya-n: There should be a way to do this before getting all the documents from the database. Something like `array-contains`

walk-in-interview/src/main/java/com/google/job/data/JobsDatabase.java
```diff
@@ -138,4 +140,57 @@
...
+                    jobs.add(job);
+                }
+            }
+
+            return jobs.build();
+        };
+
+        return ApiFutures.transform(querySnapshotFuture, function, MoreExecutors.directExecutor());
+
+//        // Runs an asynchronous transaction
```
riya-n, line 171: don't check in commented code

walk-in-interview/src/main/java/com/google/job/data/JobsDatabase.java
SibingWu: i tried to use `array-contains`, but this is `OR` operation, what i need is `AND` operation😂
and also cloud firestore can only make `requirement containsAll skills`, but what i need is `skills containsAll requirement`

htthach: Should we change array to dictionary?
https://stackoverflow.com/questions/54987399/firestore-search-array-contains-for-multiple-values

walk-in-interview/src/test/java/com/google/job/data/JobsDatabaseTest.java
```diff
@@ -245,6 +246,310 @@ public void fetchJob_normalInput_success() throws ExecutionException, Interrupte
...
+        JobStatus jobStatus = JobStatus.ACTIVE;
+        String jobName = "Programmer";
+        Location location =  new Location("Maple Tree", "123456", 0, 0);
+        String jobDescription = "Fighting to defeat hair line recede";
+        JobPayment jobPayment = new JobPayment(0, 5000, PaymentFrequency.MONTHLY);
+        List<String> requirements = Requirement.getLocalizedNames(Arrays.asList(O_LEVEL), "en");
+        long postExpiry = System.currentTimeMillis();;
+        JobDuration jobDuration = JobDuration.ONE_MONTH;
+
+        Job job1 = Job.newBuilder()
```
htthach: Maybe we can abstract test job data creation to another method? They're repeated here and a few other places.

walk-in-interview/src/main/java/com/google/job/data/Job.java
```diff
@@ -70,7 +70,7 @@ public static JobBuilder newBuilder() {
     public static final class JobBuilder {
         // Optional parameters - initialized to default values
         private String jobId = "";
-        private List<String> requirements = ImmutableList.of();
+        private Map<String, Boolean> requirements = ImmutableMap.of(); // Maps with non-string keys are not supported
```
htthach: Maybe the comment is not needed as the declaration is pretty clear?

walk-in-interview/src/main/java/com/google/job/data/Job.java
```diff
@@ -203,8 +203,8 @@ public JobPayment getJobPay() {
         return jobPay;
     }
 
-    /** Returns a list of requirements of the job post. */
-    public List<String> getRequirements() {
+    /** Returns a list of requirements (stable ids) of the job post. */
```
htthach: Nit: returns a map.

walk-in-interview/src/main/java/com/google/job/data/JobsDatabase.java
```diff
@@ -132,4 +134,41 @@
...
+        final Query activeJobsQuery = FireStoreUtils.getFireStore()
+                .collection(JOB_COLLECTION)
+                .whereEqualTo(JOB_STATUS_FIELD, JobStatus.ACTIVE);
+
+        // Eligible post: post whose requirements do not contain (field is false)
+        // the skills that applicant does not have
+        Query eligiblePostQuery = activeJobsQuery;
+        for (String negateSkill: negateSkills) {
+            String fieldPath = String.format("%s.%s", JOB_REQUIREMENTS_FIELD, negateSkill);
+            eligiblePostQuery = eligiblePostQuery.whereEqualTo(fieldPath, false);
```
htthach, line 162: Would it work if the key is not inside the job?
E.g at first there are 3 possible requirements, maybe next year there are total 5 possible requirements e.g height above 1.6m, weight above 60kg (just example). Then some of the job will not has that key. Will the query consider them as false?

walk-in-interview/src/test/java/com/google/job/data/JobsDatabaseTest.java
```diff
@@ -245,6 +248,135 @@ public void fetchJob_normalInput_success() throws ExecutionException, Interrupte
         assertEquals(job, actualJob);
     }
 
+    @Test
+    public void fetchAllEligibleJobs_normalInput_success() throws IOException, ExecutionException, InterruptedException {
+        // Arrange.
+        Map<String, Boolean> requirements = ImmutableMap.of(DRIVING_LICENSE_C.getRequirementId(), false,
+                O_LEVEL.getRequirementId(), true, ENGLISH.getRequirementId(), false);
```
htthach: Maybe some of the test job have less than 3 keys? e.g { english: true, o_level:true }

walk-in-interview/src/main/java/com/google/job/data/Job.java
```diff
@@ -14,7 +14,7 @@
     private final Location jobLocation;
     private final String jobDescription;
     private final JobPayment jobPay;
-    private final List<String> requirements;
+    private final Map<String, Boolean> requirements; // requirement stable id : true/false
```
htthach, line 17: If a key is mapped to false, does the key need to be there?

walk-in-interview/src/main/java/com/google/job/data/JobsDatabase.java
```diff
@@ -138,54 +137,38 @@
 
     /** Returns future of all ACTIVE and eligible job posts in database. */
     public Future<Collection<Job>> fetchAllEligibleJobs(List<String> skills) throws IOException {
+        // Gets all requirements stable id
+        List<String> requirementsList = Requirement.getAllRequirementIds();
+
+        // Gets a list of requirements which the applicant does not have
+        List<String> negateSkills = new ArrayList<>(requirementsList);
```
riya-n: nit: thinking if there's a better name for this...`excludedSkills`?

walk-in-interview/src/main/java/com/google/job/data/Requirement.java
```diff
@@ -26,6 +27,22 @@ public String getRequirementId() {
         return requirementId;
     }
 
+    /** Gets the requirement stable ids given enum value. */
+    public static List<String> getRequirementIds(List<Requirement> requirements) {
```
riya-n, line 31: can this be private?

riya-n: also, why is this method necessary, isn't getRequirementId() enough

walk-in-interview/src/main/java/com/google/job/data/Requirement.java
```diff
@@ -56,4 +73,13 @@ public String getLocalizedName(String language) throws IllegalArgumentException
 
         return localizedNames.build();
     }
+
+    /**
+     * Returns the localized names of the all requirements.
+     *
+     * @throw IllegalArgumentException If the language is not supported.
+     */
+    public static List<String> getAllLocalizedNames(String language) throws IllegalArgumentException {
```
riya-n, line 82: again, isn't getLocalizedName() enough?

walk-in-interview/src/test/java/com/google/job/data/JobsDatabaseTest.java
```diff
@@ -1,6 +1,7 @@
 package com.google.job.data;
 
 import com.google.api.core.ApiFuture;
+import com.google.appengine.repackaged.com.google.common.collect.ImmutableMap;
```
riya-n: wrong import? com.google.common.collect.ImmutableMap

walk-in-interview/src/test/java/com/google/job/data/JobsDatabaseTest.java
```diff
@@ -55,8 +56,8 @@ public void addJob_normalInput_success() throws ExecutionException, InterruptedE
         Location expectedLocation =  new Location("Google", "123456", 0, 0);
         String expectedJobDescription = "Programming using java";
         JobPayment expectedJobPayment = new JobPayment(0, 5000, PaymentFrequency.MONTHLY);
-        List<String> expectedRequirements = Requirement.getLocalizedNames(
-                Arrays.asList(DRIVING_LICENSE_C, O_LEVEL, ENGLISH), "en");
+        Map<String, Boolean> expectedRequirements = ImmutableMap.of(DRIVING_LICENSE_C.getRequirementId(), true,
+                O_LEVEL.getRequirementId(), true, ENGLISH.getRequirementId(), true);
```
riya-n: nit: maybe separate this on different lines so the key-value pair is clear.
```
Map<String, Boolean> expectedRequirements = ImmutableMap.of(
                                      DRIVING_LICENSE_C.getRequirementId(), true,
                                      O_LEVEL.getRequirementId(), true,
                                      ENGLISH.getRequirementId(), true);
```

riya-n: same for the ones below

walk-in-interview/src/test/java/com/google/job/data/JobsDatabaseTest.java
```diff
@@ -249,181 +251,116 @@ public void fetchJob_normalInput_success() throws ExecutionException, Interrupte
...
-                .setJobTitle(jobName)
-                .setLocation(location)
-                .setJobDescription(jobDescription)
-                .setJobPay(jobPayment)
-                .setRequirements(requirements)
-                .setPostExpiry(postExpiry)
-                .setJobDuration(jobDuration)
-                .build();
+        requirements = ImmutableMap.of(DRIVING_LICENSE_C.getRequirementId(), true,
+                O_LEVEL.getRequirementId(), true, ENGLISH.getRequirementId(), true);;
```
riya-n: nit: double semicolon

walk-in-interview/src/test/java/com/google/job/data/JobsDatabaseTest.java
```diff
@@ -249,181 +251,116 @@ public void fetchJob_normalInput_success() throws ExecutionException, Interrupte
...
+        Job job4 = testJobDataCreation(JobStatus.ACTIVE, requirements);
 
-        // future.get() blocks on response.
-        firestore.collection(TEST_JOB_COLLECTION).add(job5).get();
+        requirements = ImmutableMap.of(DRIVING_LICENSE_C.getRequirementId(), false,
+                O_LEVEL.getRequirementId(), false, ENGLISH.getRequirementId(), true);;
```
riya-n: nit: double semicolon

walk-in-interview/src/test/java/com/google/job/data/JobsDatabaseTest.java
```diff
@@ -249,181 +251,116 @@ public void fetchJob_normalInput_success() throws ExecutionException, Interrupte
...
-        // future.get() blocks on response.
-        firestore.collection(TEST_JOB_COLLECTION).add(job4).get();
+        // Assert.
+        Collection<Job> expectedJobs = new HashSet<>(Arrays.asList(job1, job5));
+        Collection<Job> actualJobs = jobsFuture.get();
+        assertEquals(expectedJobs, actualJobs);
+    }
 
-        requirements = Requirement.getLocalizedNames(Arrays.asList(ENGLISH), "en");
+    private Job testJobDataCreation(JobStatus jobStatus, Map<String, Boolean> requirements)
```
riya-n: if this is specific only to the requirements test, then maybe rename it

walk-in-interview/src/main/java/com/google/job/data/Job.java
SibingWu: since we need to find those post that does not contain skills (mapped to false) that applicant does not have (negate skills), so we need to map it to false; else cloud firestore cannot do "not in"

walk-in-interview/src/main/java/com/google/job/data/Requirement.java
SibingWu: getRequirementId() is for single Requirement, getRequirementIds() is for a specific list of requirements; will be more flexible

walk-in-interview/src/main/java/com/google/job/data/Requirement.java
SibingWu: in fact, i am considering if we need to store localized name in database... if we don't need it at all, then there is no need to localized name field for requirement; but i will leave it there for now

walk-in-interview/src/main/java/com/google/job/data/JobsDatabase.java
SibingWu: that is a problem, i will fix by adding a check to fill all requirements field at constructor

riya-n (approved):

htthach (approved):

PR #66, opened 2020-07-22
Add get individual job servlet

riya-n:
This is just the servlet to get the individual job post given the jobId. It makes use of the JobDatabase.fetchJob() function that has already been written.

### DIFF
--- a/walk-in-interview/src/main/java/com/google/job/servlets/JobServlet.java
+++ b/walk-in-interview/src/main/java/com/google/job/servlets/JobServlet.java
@@ -147,7 +147,7 @@ public final class JobServlet extends HttpServlet {
     }
 
     /**
-     * Returns the Job object.
+     * Returns the optional Job object.
      *
      * @param jobId The job id that corresponds to the job we want to get.
      * @return optional Job object with all the details of the job.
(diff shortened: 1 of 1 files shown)


SibingWu (approved):

walk-in-interview/src/main/java/com/google/job/servlets/JobServlet.java
```diff
@@ -44,7 +46,17 @@ public void service(HttpServletRequest request, HttpServletResponse response) th
 
     @Override
     public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
```
SibingWu: just make sure: here need to throw IOException? or simply add that in the catch block?

walk-in-interview/src/main/java/com/google/job/servlets/JobServlet.java
```diff
@@ -128,4 +140,42 @@ private void updateJobPost(String jobId, Job job)
...
+     * @return Job object with all the details of the job.
+     */
+    private Job fetchJobDetails(String jobId) throws ServletException, ExecutionException, TimeoutException {
+        try {
+            Optional<Job> job = this.jobsDatabase.fetchJob(jobId)
+                    .get(TIMEOUT_SECONDS, TimeUnit.SECONDS);
+
+            if (!job.isPresent()) {
+                throw new IllegalArgumentException("could not find job for this jobId: " + jobId);
+            }
```
htthach: And you can also handle in the servlet so that you return error 404 resource not found. When a client receive 404, they should know that they're requesting a resource that's not there.

walk-in-interview/src/main/java/com/google/job/servlets/JobServlet.java
riya-n: yep, added it to the catch block

htthach (approved):

PR #72, opened 2020-07-27
Fixes to small bugs noticed in the master branch

riya-n:
- updated the rendering of the requirements list given the new way in which it's stored

### DIFF
--- a/walk-in-interview/src/test/java/com/google/job/data/JobsDatabaseTest.java
+++ b/walk-in-interview/src/test/java/com/google/job/data/JobsDatabaseTest.java
@@ -21,8 +21,7 @@ public final class JobsDatabaseTest {
     // TODO(issue/15): Add failure test case
 
     private static final String TEST_JOB_COLLECTION = "Jobs";
-    // Should delete all the documents that the tests create.
-    private static final int BATCH_SIZE = 20;
+    private static final int BATCH_SIZE = 10;
 
     static JobsDatabase jobsDatabase;
     static Firestore firestore;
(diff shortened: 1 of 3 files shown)


walk-in-interview/src/test/java/com/google/job/data/JobsDatabaseTest.java
```diff
@@ -21,7 +21,8 @@
     // TODO(issue/15): Add failure test case
 
     private static final String TEST_JOB_COLLECTION = "Jobs";
-    private static final int BATCH_SIZE = 10;
+    // should delete all the documents that the tests create
```
htthach: Nit: Start with capital, end with period.

walk-in-interview/src/test/java/com/google/job/data/JobsDatabaseTest.java
```diff
@@ -21,7 +21,8 @@
...
+    private static final int BATCH_SIZE = 20;
```
htthach: As the logic of deleteCollection is that we keep deleting if count of delete >= batch size. How would large batch size ensure that we delete all documents created by test while a lower batch size does not?

SibingWu (approved):

walk-in-interview/src/test/java/com/google/job/data/JobsDatabaseTest.java
riya-n: That's true. It seems like the issue might be with the database rather than something on our side. I've reverted these changes.

htthach (approved):
Nit: please change PR description about the batch size change.

PR #74, opened 2020-07-28
Render homepage & job details map

riya-n:
Note that this does not include the Places API (so the coordinates for the jobs are still hard-coded values).
The homepage map has a marker for each job with an info window with some details about the job. The user can double click on the marker to be taken to the job details page. The job details page also includes a map, but only with the marker of that job.

This is how it looks on the homepage:
image

### DIFF
--- a/walk-in-interview/src/main/webapp/maps.js
+++ b/walk-in-interview/src/main/webapp/maps.js
@@ -1,0 +1,79 @@
+/**
+ * This file is specific to rendering and add markers to a specific map.
+ */
+
+import {SG_LATITUDE, SG_LONGITUDE, SG_MAP_ZOOM} from './common-functions.js';
+
+const jobMarkers = [];
+
+/**
+ * Creates a map in the provided html div.
+ * @param {String} mapElementId The map element id.
+ * @return {google.maps.Map} The map object created.
+ */
+function createMap(mapElementId) {
+  if (mapElementId === '') {
+    throw new Error('map element id should not be empty');
+  }
+
+  return new google.maps.Map(document.getElementById(mapElementId), {
+    center: {lat: SG_LATITUDE, lng: SG_LONGITUDE},
+    zoom: SG_MAP_ZOOM,
+  });
+}
+
+// /**
+//  * Add markers for the provided array of jobs.
+//  * The "i * 200" essentially means that each marker will be placed 200ms
+//  * after each other. It's a good value because it is short enough that the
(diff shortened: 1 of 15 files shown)


walk-in-interview/src/main/webapp/script.js
```diff
@@ -256,6 +261,22 @@ function buildJobElement(job) {
     detailsForm.submit();
   });
 
+  const marker = addMarker(map, job);
+  marker.addListener('click', function() {
+    new google.maps.InfoWindow({
+      content: jobListing.innerHTML,
```
SibingWu: * does the jobListing mean one job post listed on the home page? maybe rename to jobPost?
* can u take a screenshot of how it looks like? i am not sure if it is user-friendly to contain all info appearing on job post on homepage in a small info window, especially this web is viewed on mobile phone screen

walk-in-interview/src/main/webapp/maps.js
```diff
@@ -0,0 +1,55 @@
...
+    center: {lat: latitude, lng: longitude},
+    zoom: zoom,
+  });
+}
+
+/**
+ * Drops the markers so that the user can see them falling.
+ * The pins indicate the location of the job. You can also
+ * click on them to get some information about them.
+ * @param {google.maps.Map} map The map to add the marker to.
```
yohanlaunay, line 39: super nit: newline before.

walk-in-interview/src/main/webapp/maps.js
```diff
@@ -0,0 +1,55 @@
...
+// coordinates for centering the map around Singapore
+const SG_LATITUDE = 1.3521;
+const SG_LONGITUDE = 103.8198;
+const SG_MAP_ZOOM = 11;
+// appropriate zoom for the map to show the individual job
+const JOB_MAP_ZOOM = 15;
+
+/**
+ * Creates a map in the provided html div.
+ * @param {String} mapElementId The map element id.
```
yohanlaunay, line 16: super nit: newline before

walk-in-interview/src/main/webapp/script.js
yohanlaunay: agreed. maybe you need a job preview content.

yohanlaunay (approved):

walk-in-interview/src/main/webapp/script.js
riya-n: I renamed it to jobPostPreview. Here's what is looks like:
image

SibingWu (approved):

PR #81, opened 2020-08-03
Incorporate firebase auth into business log in front end

SibingWu:
As per title, also specify the endpoints of log in

### DIFF
--- a/walk-in-interview/src/main/webapp/common-functions.js
+++ b/walk-in-interview/src/main/webapp/common-functions.js
@@ -63,13 +63,13 @@ function renderSelectOptions(select, options) {
 /**
  * Gets the value of a specified cookie.
  * 
- * @param {String} cname 
+ * @param {String} cname Cookie name.
  */
 function getCookie(cname) {
   var name = `${cname}=`;
   var decodedCookie = decodeURIComponent(document.cookie);
   var ca = decodedCookie.split(';');
-  for(var i = 0; i <ca.length; i++) {
+  for(var i = 0; i < ca.length; i++) {
     var c = ca[i];
     while (c.charAt(0) == ' ') {
       c = c.substring(1);
@@ -78,7 +78,7 @@ function getCookie(cname) {
       return c.substring(name.length, c.length);
     }
   }
-  return "";
+  return '';
 }
 
 export {JOB_ID_PARAM, getRequirementsList, setErrorMessage,
--- a/walk-in-interview/src/main/webapp/common-functions.js
+++ b/walk-in-interview/src/main/webapp/common-functions.js
@@ -60,5 +60,26 @@ function renderSelectOptions(select, options) {
   }
 }
 
+/**
+ * Gets the value of a specified cookie.
+ * 
+ * @param {String} cname 
+ */
+function getCookie(cname) {
+  var name = `${cname}=`;
+  var decodedCookie = decodeURIComponent(document.cookie);
+  var ca = decodedCookie.split(';');
+  for(var i = 0; i <ca.length; i++) {
+    var c = ca[i];
+    while (c.charAt(0) == ' ') {
+      c = c.substring(1);
+    }
+    if (c.indexOf(name) == 0) {
+      return c.substring(name.length, c.length);
+    }
+  }
+  return "";
+}
+
 export {JOB_ID_PARAM, getRequirementsList, setErrorMessage,
-  renderSelectOptions};
+  renderSelectOptions, getCookie};
--- a/walk-in-interview/src/main/webapp/log-in/business/script.js
+++ b/walk-in-interview/src/main/webapp/log-in/business/script.js
@@ -62,7 +62,7 @@ backButton.addEventListener('click', (_) => {
 const submitButton = document.getElementById('submit');
 submitButton.addEventListener('click', (_) => {
   // Disables the button to avoid accidental double click
-  this.setAttribute('disabled', true);
+  document.getElementById('submit').setAttribute('disabled', true);
 
   const account = document.getElementById('account-input').value;
 
--- a/walk-in-interview/src/main/webapp/log-in/business/script.js
+++ b/walk-in-interview/src/main/webapp/log-in/business/script.js
@@ -62,7 +62,7 @@ backButton.addEventListener('click', (_) => {
 const submitButton = document.getElementById('submit');
 submitButton.addEventListener('click', (_) => {
   // Disables the button to avoid accidental double click
-  document.getElementById('submit').setAttribute('disabled', true);
+  document.getElementById('submit').disabled = true;
 
   const account = document.getElementById('account-input').value;
 
@@ -83,11 +83,11 @@ submitButton.addEventListener('click', (_) => {
   signIntoBusinessAccount(account, password)
       .then(() => {
         // Enables the button regardless of success or failure
-        document.getElementById('submit').setAttribute('disabled', false);
+        document.getElementById('submit').disabled = false;
       })
       .catch((error) => {
         showErrorMessageFromError(error);
-        document.getElementById('submit').setAttribute('disabled', false);
+        document.getElementById('submit').disabled = false;
       });
 });
 
--- a/walk-in-interview/src/main/webapp/log-in/business/script.js
+++ b/walk-in-interview/src/main/webapp/log-in/business/script.js
@@ -12,10 +12,16 @@ const CurrentLocale = 'en';
  */
 import {AppStrings} from '../../strings.en.js';
 import {signIntoBusinessAccount} from '../firebase-auth.js';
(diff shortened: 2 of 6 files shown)


walk-in-interview/src/main/webapp/common-functions.js
```diff
@@ -60,5 +60,26 @@ function renderSelectOptions(select, options) {
   }
 }
 
+/**
+ * Gets the value of a specified cookie.
```
riya-n, line 64: nit: include why we need this cookie function

walk-in-interview/src/main/webapp/strings.en.js
```diff
@@ -67,6 +67,11 @@ AppStrings['log-in'] = {
 AppStrings['business-log-in'] = {
   'account': 'Account (Email)',
   'password': 'Password',
+  'empty-email': 'Empty email input',
+  'invalid-email-error': 'The email address is not valid',
+  'user-disabled': 'The user corresponding to the given email has been disabled',
```
riya-n: what does disabled user mean?

walk-in-interview/src/main/webapp/log-in/business/script.js
```diff
@@ -11,11 +11,18 @@ const CurrentLocale = 'en';
...
  */
 import {AppStrings} from '../../strings.en.js';
+import {signIntoBusinessAccount} from '../firebase-auth.js';
+import {setErrorMessage} from '../../common-functions.js';
 
 const COMMONG_STRINGS = AppStrings['log-in'];
 const STRINGS = AppStrings['business-log-in'];
 const LOGIN_HOMEPAGE_PATH = '../index.html';
 
+const INVALID_EMAIL_ERROR_CODE = 'auth/invalid-email';
```
riya-n, line 21: not sure if we need constants for these

walk-in-interview/src/main/webapp/log-in/business/script.js
```diff
@@ -55,4 +62,50 @@ backButton.addEventListener('click', (_) => {
...
+                /* includesDefault= */false);
+            break;
+          case WRONG_PASSWORD_ERROR_CODE:
+            setErrorMessage(/* errorMessageElementId= */'error-message',
+                /* msg= */ STRINGS['wrong-password'],
+                /* includesDefault= */false);
+            break;
+          default:
+            setErrorMessage(/* errorMessageElementId= */'error-message',
+                /* msg= */ error.message,
```
riya-n: not sure if we should display the error msg to the user - maybe just print it in the console?

walk-in-interview/src/main/webapp/log-in/business/script.js
```diff
@@ -55,4 +62,50 @@ backButton.addEventListener('click', (_) => {
...
+  }
+
+  const password = document.getElementById('password-input').value;
+
+  setErrorMessage(/* errorMessageElementId= */'error-message',
+      /* msg= */ '',
+      /* includesDefault= */false);
+
+  signIntoBusinessAccount(account, password)
+      .catch(error => {
```
riya-n: nit: `(error)` instead of `error`

walk-in-interview/src/main/webapp/log-in/firebase-auth.js
```diff
@@ -50,11 +52,17 @@ function createBusinessAccount(email, password) {
...
  */
 function signIntoBusinessAccount(email, password) {
-  firebase.auth().signInWithEmailAndPassword(email, password)
-      .catch((error) => {
-        console.error(error);
+  return firebase.auth().signInWithEmailAndPassword(email, password)
+      .then(({user}) => {
+        // Get the user's ID token as it is needed to exchange for a session cookie.
+        return user.getIdToken()
+            .then(idToken => {
```
riya-n: nit: `(idToken)`

walk-in-interview/src/main/webapp/log-in/firebase-auth.js
```diff
@@ -104,5 +112,25 @@ function checkCurrentUser() {
   });
 }
 
+/**
+ * Makes a POST request to session log in endpoint.
+ * 
+ * @param {*} url Login endpoint.
```
riya-n: nit: include the param types?

walk-in-interview/src/main/webapp/common-functions.js
SibingWu: Screenshot 2020-08-03 at 5 01 08 PM

walk-in-interview/src/main/webapp/log-in/business/script.js
SibingWu: I will change it to a general error message, and only print the exact cause in the console

walk-in-interview/src/main/webapp/strings.en.js
SibingWu: A disabled user cannot sign in to an app anymore and continues to be stored in the Firebase Auth backend.

walk-in-interview/src/main/webapp/common-functions.js
riya-n: I think you should include something about this in the docs for the function e.g., this function is needed to help get the idToken for firebase auth ...

walk-in-interview/src/main/webapp/log-in/business/script.js
```diff
@@ -55,4 +62,51 @@ backButton.addEventListener('click', (_) => {
...
+  const account = document.getElementById('account-input').value;
+
+  if (account === '') {
+    setErrorMessage(/* errorMessageElementId= */'error-message',
+        /* msg= */ STRINGS['empty-email'],
+        /* includesDefault= */false);
+    return;
+  }
+
+  const password = document.getElementById('password-input').value;
```
riya-n, line 76: do we need to add a check for empty password field?

walk-in-interview/src/main/webapp/log-in/business/script.js
```diff
@@ -55,4 +62,51 @@ backButton.addEventListener('click', (_) => {
...
+
+  // Resets the error (there might have been an error msg from earlier)
+  setErrorMessage(/* errorMessageElementId= */'error-message',
```
riya-n, line 79: shouldn't the error message field only reset if the sign in is successful?

walk-in-interview/src/main/webapp/strings.en.js
riya-n: I think you should change the error message shown to the user so it's more clear what the error is. How does a user become disabled?

walk-in-interview/src/main/webapp/common-functions.js
SibingWu: I don't think so, since `getCookie` is a very commonly used function (that is also why i put it inside common-functions.js), just in this case, it is used only once

walk-in-interview/src/main/webapp/log-in/business/script.js
SibingWu: i think no need, since if it is empty, it is regarded as wrong password

walk-in-interview/src/main/webapp/log-in/business/script.js
SibingWu: it is just reset to make sure it is empty before starting signing in; if there is error, the error message will display again

walk-in-interview/src/main/webapp/log-in/business/script.js
```diff
@@ -55,4 +62,51 @@ backButton.addEventListener('click', (_) => {
 const submitButton = document.getElementById('submit');
 submitButton.addEventListener('click', (_) => {
```
htthach, line 63: Do we need to temporarily disable this button while logging in to avoid clicking multiple times?

walk-in-interview/src/main/webapp/log-in/business/script.js
```diff
@@ -55,4 +62,51 @@ backButton.addEventListener('click', (_) => {
...
+  const password = document.getElementById('password-input').value;
+
+  // Resets the error (there might have been an error msg from earlier)
+  setErrorMessage(/* errorMessageElementId= */'error-message',
+      /* msg= */ '',
+      /* includesDefault= */false);
+
+  signIntoBusinessAccount(account, password)
+      .catch((error) => {
+        switch(error.code) {
```
htthach: Maybe abstract these to another method? e.g showErrorMessageFromCode(errorCode) ?

htthach (approved):

riya-n (approved):

walk-in-interview/src/main/webapp/log-in/business/script.js
SibingWu: As discussed offline, disable the button before line 65, enable the button within the block of success case and failure case

htthach (approved):

Issue #56, opened 2020-07-17
Add `int` validation check for `max` for payment


SibingWu:
As per title, add check in both client and server side

PR #83, opened 2020-08-03
Integer check back end

SibingWu:
Related issue: #56 

Add integer check for job payment constructor in the back end, so that a job will not have too large max payment

### DIFF
--- a/walk-in-interview/src/main/java/com/google/job/data/JobPayment.java
+++ b/walk-in-interview/src/main/java/com/google/job/data/JobPayment.java
@@ -28,7 +28,7 @@ public final class JobPayment {
         }
 
         if (max > Integer.MAX_VALUE) {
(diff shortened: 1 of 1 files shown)


riya-n (approved):

walk-in-interview/src/main/java/com/google/job/data/JobPayment.java
```diff
@@ -27,6 +27,10 @@ public JobPayment(int min, int max, PaymentFrequency paymentFrequency) throws Il
             throw new IllegalArgumentException("\"max\" should not be less than \"min\"");
         }
 
+        if (max > Integer.MAX_VALUE) {
+            throw new IllegalArgumentException("\"max\" should be no greater than 2147483647");
```
yohanlaunay: nit: should not be greater than + Integer.MAX_VALUE

yohanlaunay (approved):

PR #93, opened 2020-08-05
Add servlet for getting interested jobs

riya-n:
- only the GET servlet has been added
- the method to fetch and iterate through the jobIds from the applicant's jobsList is yet to be implemented

### DIFF
--- a/walk-in-interview/src/main/java/com/google/job/data/JobsDatabase.java
+++ b/walk-in-interview/src/main/java/com/google/job/data/JobsDatabase.java
@@ -254,7 +254,7 @@ public final class JobsDatabase {
         return ApiFutures.transform(
             docRef.get(),
             documentSnapshot -> {
-                // TODO(issue/92): get the applican't jobsList and iterate through it to get the job documents
+                // TODO(issue/92): get the applicant's jobsList and iterate through it to get the job documents
                 // for now just return an empty job page
                 return new JobPage(/* jobList= */ ImmutableList.of(), /* totalCount= */ 0, Range.between(0, 0));
             },
--- a/walk-in-interview/src/main/java/com/google/job/data/JobsDatabase.java
+++ b/walk-in-interview/src/main/java/com/google/job/data/JobsDatabase.java
@@ -239,15 +239,13 @@ public final class JobsDatabase {
     /**
      * Gets all the applicant's interested jobs given the params.
      *
+     * @param applicantId The applicant's userId.
      * @param pageSize The the number of jobs to be shown on the page.
      * @param pageIndex The page number on which we are at.
      * @return Future of the JobPage object.
      */
-    public static Future<JobPage> fetchInterestedJobPage(int pageSize, int pageIndex) throws IOException {
+    public static Future<JobPage> fetchInterestedJobPage(String applicantId, int pageSize, int pageIndex) throws IOException {
         CollectionReference applicantAccountsCollection = FireStoreUtils.getFireStore().collection(APPLICANT_ACCOUNTS_COLLECTION);
-
-        // TODO(issue/91): get userId from firebase session cookie
-        String applicantId = "";
 
         DocumentReference docRef = applicantAccountsCollection.document(applicantId);
 
--- a/walk-in-interview/src/main/java/com/google/job/data/JobsDatabase.java
+++ b/walk-in-interview/src/main/java/com/google/job/data/JobsDatabase.java
@@ -12,6 +12,7 @@ import org.checkerframework.checker.nullness.compatqual.NullableDecl;
 import org.apache.commons.lang3.Range;
 import java.lang.UnsupportedOperationException;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeoutException;
 
 import java.io.IOException;
 import java.util.*;
--- a/walk-in-interview/src/main/java/com/google/job/data/JobsDatabase.java
+++ b/walk-in-interview/src/main/java/com/google/job/data/JobsDatabase.java
@@ -274,7 +274,16 @@ public final class JobsDatabase {
 
                 List<Job> jobList = ImmutableList.of();
                 try {
-                    jobList = fetchJobsFromIds(interestedList).get(TIMEOUT_SECONDS, TimeUnit.SECONDS);
+                    int counter = 0;
+                    while (counter + 10 < interestedList.size()) {
+                        List<Job> subList = interestedList.subList(/* inclusive */ counter, /* exclusive */ counter + 10);
+                        List<Job> fetchedList = fetchJobsFromIds(subList).get(TIMEOUT_SECONDS, TimeUnit.SECONDS);
+                        jobList.addAll(fetchedList);
+                        counter = counter + 10;
+                    }
+                    List<Job> subList = interestedList.subList(/* inclusive */ counter, /* exclusive */ interestedList.size());
+                    List<Job> fetchedList = fetchJobsFromIds(subList).get(TIMEOUT_SECONDS, TimeUnit.SECONDS);
+                    jobList.addAll(fetchedList);
                     log.info("returned jobList: " + jobList.toString());
                     // TODO(issue/34): adjust range/total count based on pagination
                 } catch (IOException | InterruptedException | ExecutionException | TimeoutException e) {
--- a/walk-in-interview/src/main/java/com/google/job/data/JobsDatabase.java
+++ b/walk-in-interview/src/main/java/com/google/job/data/JobsDatabase.java
@@ -34,12 +34,9 @@ public final class JobsDatabase {
     private static final String JOB_STATUS_FIELD = "jobStatus";
     private static final String JOB_REQUIREMENTS_FIELD = "requirements";
     private static final String INTERESTED_JOBS_FIELD = "interestedJobs";
-<<<<<<< HEAD
     private static final String JOB_ID_FIELD = "jobId";
     
     private static final long TIMEOUT_SECONDS = 5;
-=======
->>>>>>> master
 
     /**
      * Adds a newly created job post.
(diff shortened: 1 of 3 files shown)


SibingWu (approved):

walk-in-interview/src/main/java/com/google/job/data/JobsDatabase.java
```diff
@@ -233,4 +235,30 @@
...
+     *
+     * @param pageSize The the number of jobs to be shown on the page.
+     * @param pageIndex The page number on which we are at.
+     * @return Future of the JobPage object.
+     */
+    public static Future<JobPage> fetchInterestedJobPage(int pageSize, int pageIndex) throws IOException {
+        CollectionReference applicantAccountsCollection = FireStoreUtils.getFireStore().collection(APPLICANT_ACCOUNTS_COLLECTION);
+
+        // TODO(issue/91): get userId from firebase session cookie
+        String applicantId = "";
```
yohanlaunay: You should pass this as an argument instead. The jobs db doesn't care about session data, that's under purview of the servlet.

walk-in-interview/src/main/java/com/google/job/data/JobsDatabase.java
```diff
@@ -233,4 +235,30 @@
...
+
+        DocumentReference docRef = applicantAccountsCollection.document(applicantId);
+
+        return ApiFutures.transform(
+            docRef.get(),
+            documentSnapshot -> {
+                // TODO(issue/92): get the applicant's jobsList and iterate through it to get the job documents
+                // for now just return an empty job page
+                return new JobPage(/* jobList= */ ImmutableList.of(), /* totalCount= */ 0, Range.between(0, 0));
```
yohanlaunay: why not read from the snapshot? you don't need a todo here. since your document is empty you should return empty list with totalCount0.

walk-in-interview/src/main/java/com/google/job/servlets/InterestedJobsServlet.java
```diff
@@ -0,0 +1,63 @@
...
+    /**
+     * Returns the JobPage object.
+     *
+     * @param pageSize The the number of jobs to be shown on the page.
+     * @param pageIndex The page number on which we are at.
+     * @return JobPage object with all the details for the GET response.
+     */
+    private JobPage fetchJobPageDetails(int pageSize, int pageIndex) throws ServletException, ExecutionException, TimeoutException {
+        try {
+            return this.jobsDatabase.fetchInterestedJobPage(pageSize, pageIndex)
```
yohanlaunay: here: 
```
//TODO(issue/) get the applicant Id from session data.
String applicantId = "";
return [...]
```

walk-in-interview/src/main/java/com/google/job/data/JobsDatabase.java
```diff
@@ -233,4 +240,44 @@
...
+
+        return ApiFutures.transform(
+            docRef.get(),
+            documentSnapshot -> {
+                ImmutableList.Builder<Job> jobListBuilder = ImmutableList.builder();
+                List<String> interestedList = (List<String>) documentSnapshot.get("interestedJobs");
+
+                for (String jobId: interestedList) {
+                    try {
+                        Optional<Job> job = fetchJob(jobId).get(5, TimeUnit.SECONDS);
```
yohanlaunay: you are doing what is called as a N+1 select. Meaning you do a select, then get a list, and for each item in the list you do another select. Are you able to batch select all the jobs instead by passing a list of Ids? (similar to a JOIN in SQL)?

https://stackoverflow.com/questions/97197/what-is-the-n1-selects-problem-in-orm-object-relational-mapping#:~:text=The%20N%2B1%20query%20problem,the%20larger%20the%20performance%20impact.

walk-in-interview/src/main/java/com/google/job/data/JobsDatabase.java
```diff
@@ -233,4 +240,44 @@
...
+    public Future<JobPage> fetchInterestedJobPage(String applicantId, int pageSize, int pageIndex) throws IOException {
+        CollectionReference applicantAccountsCollection = FireStoreUtils.getFireStore().collection(APPLICANT_ACCOUNTS_COLLECTION);
+
+        DocumentReference docRef = applicantAccountsCollection.document(applicantId);
+
+        return ApiFutures.transform(
+            docRef.get(),
+            documentSnapshot -> {
+                ImmutableList.Builder<Job> jobListBuilder = ImmutableList.builder();
+                List<String> interestedList = (List<String>) documentSnapshot.get("interestedJobs");
```
yohanlaunay: i think it might be null here can you double check?

walk-in-interview/src/main/java/com/google/job/data/JobsDatabase.java
```diff
@@ -233,4 +240,44 @@
...
+
+                for (String jobId: interestedList) {
+                    try {
+                        Optional<Job> job = fetchJob(jobId).get(5, TimeUnit.SECONDS);
+                        if (job.isPresent()) {
+                            // Only add the job to this list if its active or valid.
+                            jobListBuilder.add(job.get());
+                        }
+                    } catch (IOException | InterruptedException | ExecutionException | TimeoutException e) {}
```
yohanlaunay: always log the error. Also document (in your error handling and javadoc) what happens when you cannot fetch a job that is in the user's favorite.

walk-in-interview/src/main/java/com/google/job/data/JobsDatabase.java
```diff
@@ -233,4 +240,44 @@
...
+        return ApiFutures.transform(
+            docRef.get(),
+            documentSnapshot -> {
+                ImmutableList.Builder<Job> jobListBuilder = ImmutableList.builder();
+                List<String> interestedList = (List<String>) documentSnapshot.get("interestedJobs");
+
+                for (String jobId: interestedList) {
+                    try {
+                        Optional<Job> job = fetchJob(jobId).get(5, TimeUnit.SECONDS);
+                        if (job.isPresent()) {
```
yohanlaunay: job must be present AND not in DELETED state. You also have to decide what to do with jobs that are in non-published state (meaning not publicly available anymore) here you are returning them so the client-side will have to filter them out OR gray them out somehow on the screen.

walk-in-interview/src/main/java/com/google/job/servlets/InterestedJobsServlet.java
```diff
@@ -0,0 +1,65 @@
...
+            int pageSize = JobsListingsServlet.parsePageIndex(request);
+            int pageIndex = JobsListingsServlet.parsePageIndex(request);
+            
+            JobPage jobPage = fetchJobPageDetails(pageSize, pageIndex);
+
+            String json = ServletUtils.convertToJsonUsingGson(jobPage);
+            response.setContentType("application/json;");
+            response.getWriter().println(json);
+        } catch(IllegalArgumentException | ServletException | ExecutionException | TimeoutException e) {
+            response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
```
yohanlaunay: can you log the exception? here it's swallowed. specially the servlet or execution exception, we really want to know when the error comes from the server-side.

walk-in-interview/src/main/java/com/google/job/data/JobsDatabase.java
riya-n: I added a check to make sure the job is in ACTIVE state

walk-in-interview/src/main/java/com/google/job/data/JobsDatabase.java
```diff
@@ -238,6 +248,99 @@
...
+                    throw new IllegalArgumentException("Invalid applicantId");
+                }
+
+                List<String> interestedList = (List<String>) documentSnapshot.get(INTERESTED_JOBS_FIELD);
+                // TODO(issue/34): adjust the interestedList the be looked at based on pagination
+
+                ImmutableList.Builder<Job> jobListBuilder = ImmutableList.builder();
+                try {
+                    int counter = 0;
+                    // Handles the jobIds 10 at a time.
```
yohanlaunay: Because the 10 is due to a restriction on firestore IN query i would actually suggest you move that logic of splitting your list inside the JobsDb manager instead. so you pass the whole list to it and the db manager will break it up in batches of 10 and return the combined list.

Also replace the 10 (magic number) by a constant so you don't need to explain what it is,m something like FIRESTORE_IN_QUERY_MAX_ARGS

walk-in-interview/src/main/java/com/google/job/data/JobsDatabase.java
```diff
@@ -238,6 +249,109 @@
...
+    /**
+     * This will return a future of a list of jobs given the list of jobIds.
+     * Any jobs that are not active or invalid will not be included in the list returned.
+     * This only handles jobIds of max 10 at a time.
+     *
+     * @param jobIds The list of jobIds.
+     * @return Future of the list of jobs.
+     */
+    private Future<List<Job>> fetchJobsFromIds(List<String> jobIds) throws IOException {
+        CollectionReference jobsCollection = FireStoreUtils.getFireStore().collection(JOB_COLLECTION);
```
yohanlaunay, line 333: FYI you should check if the list is empty and immediately return empty list back.

walk-in-interview/src/main/java/com/google/job/data/JobsDatabase.java
```diff
@@ -238,6 +249,109 @@
...
+            log.log(Level.SEVERE, "error while getting interested job list ", e);
+        }
+
+        return jobListBuilder.build();
+    }
+
+    /**
+     * This will return a future of a list of jobs given the list of jobIds.
+     * Any jobs that are not active or invalid will not be included in the list returned.
+     * This only handles jobIds of max 10 at a time.
```
yohanlaunay: The 10 is defined as constant so it might change you should use `This can only handle up to {@link #FIRESTORE_IN_QUERY_MAX_ARGS} jobIds at a time`

yohanlaunay (approved):

PR #97, opened 2020-08-07
Fix bug in master

SibingWu:
Fixed some major bug(which stops the program) found when running the master branch


riya-n (approved):

htthach (approved):

walk-in-interview/src/main/java/com/google/job/data/JobsDatabase.java
```diff
@@ -326,10 +328,6 @@
      * @return Future of the list of jobs.
      */
     private Future<List<Job>> fetchJobsFromIds(List<String> jobIds) throws IOException {
-        if (jobIds.isEmpty()) {
-            return ImmutableList.of();
```
yohanlaunay, line 330: you want:
```
if( jobIds.isEmpty() ){
   return Futures.immediateFuture(ImmutableList.of());
}
```

PR #112, opened 2020-08-19
Create business account back end

SibingWu:
* Implement the backend for create business account feature
* Create the `Business` object
* Move the creating preliminary stuff to the backend
* Write tests for `getBusinessAccount` and `updateBusinessAccount`


walk-in-interview/src/main/java/com/google/account/business/servlets/CreateBusinessServlet.java
```diff
@@ -0,0 +1,96 @@
...
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.stream.Collectors;
+
+/** Servlet that handles creating new business account. */
+@WebServlet("/business-account")
+public final class CreateBusinessServlet extends HttpServlet {
+    private static final long TIMEOUT_SECONDS = 5;
+    private BusinessDatabase businessDatabase;
```
yohanlaunay, line 32: nit: newline between static and non-static

walk-in-interview/src/main/java/com/google/account/business/servlets/CreateBusinessServlet.java
```diff
@@ -0,0 +1,96 @@
...
+        this.businessDatabase = new BusinessDatabase();
+    }
+
+    @Override
+    public void doPost(HttpServletRequest request, HttpServletResponse response) {
+        try {
+            Optional<String> optionalUid = FirebaseAuthUtils.getUid(request);
+
+            if (!optionalUid.isPresent()) {
+                System.err.println("Illegal uid");
```
yohanlaunay: nit: i thought you had another logger? can you use that one instead of System?

walk-in-interview/src/main/java/com/google/account/business/servlets/CreateBusinessServlet.java
```diff
@@ -0,0 +1,96 @@
...
+            Business rawBusiness = parseBusinessAccount(request);
+
+            // Forces new account to have empty job list
+            Business business = rawBusiness.toBuilder().setJobs(ImmutableList.of()).build();
+
+            // Stores the account into cloud firestore
+            addBusinessAccount(uid, business);
+
+            // Sends the success status code in the response
+            response.setStatus(HttpServletResponse.SC_OK);
```
yohanlaunay, line 76: any reason why you don't return the session cookie too here?

walk-in-interview/src/main/java/com/google/account/UserType.java
```diff
@@ -0,0 +1,18 @@
...
+    BUSINESS("BUSINESS"),
+    APPLICANT("APPLICANT");
+
+    private final String userTypeId;
+
+    UserType(String userTypeId) {
+        this.userTypeId = userTypeId;
+    }
+
+    /** Returns the stable id representing the job duration. Can be stored in database. */
```
htthach: Nit: I think we have a wrong comment here.

walk-in-interview/src/main/java/com/google/account/business/data/Business.java
```diff
@@ -0,0 +1,139 @@
...
+        this.jobs = ImmutableList.of();
+    }
+
+    private Business(BusinessBuilder businessBuilder) {
+        this.userType = businessBuilder.userType;
+        this.name = businessBuilder.name;
+        this.jobs = businessBuilder.jobs;
+    }
+
+    /** Returns a builder by copying all the fields of an existing Job. */
```
htthach: Nit: I think the comment has wrong content, "existing Business" ?

walk-in-interview/src/main/java/com/google/account/business/servlets/CreateBusinessServlet.java
```diff
@@ -0,0 +1,96 @@
...
+            // Validates the attributes via build()
+            return rawBusiness.toBuilder().build();
+        }
+    }
+
+    /** Adds the business account in the database. */
+    private void addBusinessAccount(String uid, Business business)
+            throws IllegalArgumentException, ServletException, ExecutionException, TimeoutException {
+        try {
+            // Blocks the operation.
```
htthach: What do you mean by "Blocks the operation" ? Did you mean "This is a blocking operation" ?

walk-in-interview/src/main/java/com/google/account/business/data/BusinessDatabase.java
```diff
@@ -0,0 +1,21 @@
...
+import com.google.utils.FireStoreUtils;
+
+import java.io.IOException;
+import java.util.concurrent.Future;
+
+/** Helps persist and retrieve business accounts in cloud firestore. */
+public final class BusinessDatabase {
+    private static final String BUSINESS_ACCOUNT_COLLECTION = "BusinessAccounts";
+
+    // Creates a new business object and stores into the cloud firestore.
```
htthach: Nit: maybe use /** */ for java doc.

walk-in-interview/src/test/java/com/google/account/business/data/BusinessDatabaseTest.java
```diff
@@ -0,0 +1,101 @@
...
+        future.get();
+
+        DocumentReference documentReference = FireStoreUtils.getFireStore()
+                .collection(TEST_BUSINESS_COLLECTION).document(uid);
+        DocumentSnapshot documentSnapshot = documentReference.get().get();
+
+        Business actualBusiness = documentSnapshot.toObject(Business.class);
+
+        assertEquals(expectedBusiness, actualBusiness);
+    }
```
htthach, line 108: Maybe another test that create multiple account? and see if we have multiple?
There could be a bug that we overwrite existing collection when adding new account?
And another test that create duplicated business?

walk-in-interview/src/main/java/com/google/account/business/data/BusinessDatabase.java
```diff
@@ -0,0 +1,21 @@
...
+    public Future<WriteResult> createBusinessAccount(String uid, Business business) throws IOException {
```
htthach: Is this create or update?
If create, can we create two business with the same id? Would the future fail?

SibingWu: if with the same uid, the later one will overlap the first one

walk-in-interview/src/test/java/com/google/account/business/data/BusinessDatabaseTest.java
SibingWu: there should not be such bug, since the document id of the account is the uid of the account, while firebase auth will generate only unique uid for each account

and what do u mean by "duplicated business"?

htthach: Since we now changed from "create" to "update" so we probably don't need to test duplicated create. Aka does the database allow createItem(duplicatedId) multiple times.

htthach (approved):

walk-in-interview/src/main/java/com/google/account/business/servlets/CreateBusinessServlet.java
```diff
@@ -0,0 +1,114 @@
...
+            Business business;
+
+            // Updates the preliminary business object at only the updatable fields
+            if (!existingBusinessOptional.isPresent()) {
+                business = Business.newBuilder()
+                        .setName(inputBusiness.getName())
+                        .setJobs(ImmutableList.of()).build();
+            } else {
+                business = Business.newBuilder()
+                        .setName(inputBusiness.getName())
```
yohanlaunay, line 68: does that mean that only set fields will be updated? the others will remain untouched?

A better way might be:
```
business = existingBusinessOptional.toBuilder().setName(...).build();
```

note the "toBuilder" not "newBuilder".

OR even better better way is to NOT update the full business object but only the field.
Basically directly updating the name field in firebase instead of setting a name field to an empty business object and persisting the business object.

walk-in-interview/src/main/java/com/google/account/business/servlets/CreateBusinessServlet.java
```diff
@@ -0,0 +1,114 @@
...
+
+            Business rawBusiness = ServletUtils.parseFromJsonUsingGson(businessAccountJsonStr, Business.class);
+
+            // Validates the attributes via build()
+            return rawBusiness.toBuilder().build();
+        }
+    }
+
+    /** Adds the business account in the database. */
+    private void addBusinessAccount(String uid, Business business)
```
yohanlaunay: rename: updateBusinessAccount

walk-in-interview/src/main/java/com/google/account/business/servlets/CreateBusinessServlet.java
```diff
@@ -0,0 +1,114 @@
...
+            }
+
+            Business rawBusiness = ServletUtils.parseFromJsonUsingGson(businessAccountJsonStr, Business.class);
+
+            // Validates the attributes via build()
+            return rawBusiness.toBuilder().build();
+        }
+    }
+
+    /** Adds the business account in the database. */
```
yohanlaunay: change javadoc to reflect that existing business will be updated and only the fields that are set.

yohanlaunay (approved):

